PR #10, opened 2015-01-03
Use u32 to represent variant tags.

bitonic:
This makes it possible to exchange binary serialized structures between
different architectures, and reduces bandwidth.  I doubt it's even
possible to have more than 4 billions variants anyway :).

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -76,7 +76,8 @@ impl<'a, R: Reader+Buffer> Decoder<IoError> for DecoderReader<'a, R> {
         }
     fn read_enum_variant<T, F>(&mut self, names: &[&str], mut f: F) -> IoResult<T> where
         F: FnMut(&mut DecoderReader<'a, R>, uint) -> IoResult<T> {
-            let id = try!(self.read_uint());
+            let id = try!(self.read_u32());
+            let id = id as uint;
             if id >= names.len() {
                 Err(IoError {
                     kind: OtherIoError,
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -1,5 +1,6 @@
 use std::io::{Writer, IoError, IoResult};
 use rustc_serialize::Encoder;
+use std::num::Int;
 
 type EwResult = IoResult<()>;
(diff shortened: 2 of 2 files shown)


TyOverby:
Right now my code works between architectures because self.read_uint() and self.emit_uint() are producing u64 on both x86 and x86-64.  However, I do buy your argument that 32-bits are more than enough.  I'll merge this when I figure out what is happening  with the Travis-CI build.

TyOverby:
Actually, what about forcing emit_uint() to fit in a u32 instead of a u64?  I'll need to bump the version number because it is a breaking change, but it will make it so that once a message is encoded somewhere, it can be decoded anywhere.

bitonic:
I hadn't noticed that you use `write_be_u64` for `uint`.  I think that's a bad idea, I would use `write_uint` and leave the responsibility to the user -- code that works with machine dependent types are always going to have different semantics on different machines.

On the other hand, I think that forcing `uint` to `u32` is an even worse idea, because it's unexpected and without any good reason.

The build might be failing because you're depending on an outdated `rustc-serialize`.

TyOverby:
The thing is that uint shows up a lot in rust data structures like Vec and HashMap, so it's not really their fault.  I could do the checking on the deserialization side as an alternative.  If it tries to deserialize a value that is to big to fit in a 32-bit uint, it could fail.

TyOverby:
Also, yeah, the build was failing because I was using an old version of `rustc-serialize` and because rustc broke `derive(RustcSerialize)` unless you have a flag enabled.

bitonic:
Mhm, right, I guess serializing `uint` in a machine dependent way could be problematic if a lot of established data structures use it.  In any case, I think falling back to `u64` is a better choice, if we have to fall back to something.  I just checked and the `cereal` and `binary` Haskell libraries do the same.

What's certain is that encoding variant tags to `u32` is a win :).

TyOverby:
Thanks for the pull request, I've thought about it for a bit and decided to merge!

PR #14, opened 2015-01-07
More robust and precise error handling.

bitonic:
When decoding, decoding errors are now isolated from `IoError`s.

Decoding is also more strict in various circumstances, see new test.


bitonic:
Rebased.  I see nothing wrong with using `try!` that way -- that `FromError` trait seems to serve that purpose well.  If you want to change the style using some other function, feel free to do it and merge.

TyOverby:
Ok, I'll do it post-merge.

PR #23, opened 2015-03-02
Don't require BufRead, only Read

jmesmon:
While BufRead is preferable to avoid dog-slow performance, requiring
when no methods are called is also problematic, especially for wrappers
around Read that shouldn't necessarily need to know they're wrapping
BufRead.


TyOverby:
Oh yeah, this was an artifact from the old `Reader` trait.  Thanks for noticing, and thanks for the pull request!

PR #28, opened 2015-03-31
Fix for upstream rustc changes

mikedilger:
Fixes new rustc 2015-03-30 issue:

````
src/reader.rs:126:22: 126:42 error: the type of this value must be known in this context
src/reader.rs:126         self.read += cast(count).unwrap();
````


TyOverby:
Thanks!

PR #31, opened 2015-04-03
refbox: add Debug impl

jmesmon:
Allows `#[derive(Debug)]` to work with things containing RefBox.

Not sure adding impls like this is sustainable.

I used to be able to do
```
impl<'a> fmt::Debug for bincode::RefBox<'a, MyLocalType> {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> Result<(), fmt::Error> {
        write!(fmt, "RefBox({:?})", *self)
    }
}
```
Locally, but recent rust has forbidden this.

### DIFF
--- a/src/refbox.rs
+++ b/src/refbox.rs
@@ -1,5 +1,6 @@
 use std::boxed::Box;
 use std::ops::Deref;
+use std::fmt;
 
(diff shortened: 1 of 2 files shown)


TyOverby:
Not implementing Debug was a mistake on my part.  Thanks!

PR #32, opened 2015-04-27
Fix rustdoc comment syntax

mohtar:
The correct syntax for module level documentation is `//!`, not `///!`. This will make the doc text appear on the crate documentation's first page.

I also moved the comment block to the top of the file so that rustdoc won't throw "expected outer comment" error.


TyOverby:
Thanks!

PR #40, opened 2015-05-28
SliceBox: allow Deref into &[T]

jmesmon:
Allows more convenient use of SliceBox where one doesn't care whether
it represents a Vec<T> or &[T] and can handle using a &[T].

### DIFF
--- a/src/refbox.rs
+++ b/src/refbox.rs
@@ -322,3 +322,14 @@ impl <'a, T> Deref for RefBox<'a, T> {
         }
     }
 }
+
+impl <'a, T> Deref for SliceBox<'a, T> {
+    type Target = [T];
+
+    fn deref(&self) -> &[T] {
+        match &self.inner {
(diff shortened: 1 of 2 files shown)


TyOverby:
How are you using this in your code?  The reason that I didn't add this impl is because slicebox should be used only for immediate encoding / decoding and shouldn't really be used outside of that, so there shouldn't be a case where you aren't 100% certain that it represents a Vec or &[T].

jmesmon:
I'm using structs I encode/decode directly for things in my programs other than encoding and decoding.
Rather than create duplicates of all my encoded/decoded structs just to push the same data around, doing this was more convenient.

Note that there is already a (nearly identical) implementation of Deref for RefBox (which this code was based on) which, as far as I can see, is provided for similar uses.

TyOverby:
Fair enough; thanks for the pull request!

Issue #56, opened 2016-02-28
Crates.io build failing due to serde update


tikue:
It looks like 0.4.0 doesn't list a version range for serde, causing builds to [fail](https://travis-ci.org/google/tarpc/jobs/112377008) because Cargo is pulling in version 0.7.0.

TyOverby:
Fixed in 0.4.1

Issue #55, opened 2016-02-28
Support serde 0.7.X


tikue:
New release made some breaking changes, and it'd be great to have support.

andruhon:
+1

maciejkula:
+1

It might make sense to change the requirements to nail the serde version.

TyOverby:
on it

PR #57, opened 2016-02-28
Temporary serde dependency fix

Auruss:
As stated in #55 #56 builds are failing right now because of an update of serde.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ description = "A binary serialization / deserialization strategy and implementat
 rustc-serialize = "0.3.*"
 byteorder = "0.4.*"
 num = "^0.1"
-serde = "^0.6"
+serde = "0.6"
 
 [dev-dependencies]
 serde_macros = "*"


larsbergstrom:
r? @TyOverby 

Sorry to be pushy, but broken bincode on nightly is causing rippling CI failures across the entirety of Servo's dependency chain that uses it (many, many modules) and prevents us from landing any new PRs in those components. Thanks!

tikue:
Shouldn't ^0.6 already prevent 0.7 from being used?

Auruss:
@tikue I think ^ allows minor version updates.

tikue:
I could be wrong, but I thought only when it's not a semver breaking
update, and for versions less than 0, 0.X => 0.Y is breaking but 0.X.Y =>
0.X.Z is not breaking.

If that's the case, my best guess for what's happening is that the version
on crates.io was published before the caret restriction was added.

TyOverby:
I'll get on this.  Do the tests work for you or is the CI broken?

tikue:
Tests are mostly fine, all visit_* methods on Serialize/Deserialize impls
change to serialize_*/Deserialize_* respectively. About 100 LOC diff.

TyOverby:
I needed to update a few more things in the Cargo.toml, so I made this change and committed it.

larsbergstrom:
@TyOverby Thanks a ton!

PR #60, opened 2016-02-29
Support serde 0.7

### LINKED ISSUE #55: Support serde 0.7.X

TyOverby:
Closes #55

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bincode"
-version = "0.4.1"
+version = "0.5.0"
 authors = ["Ty Overby <ty@pre-alpha.com>", "Francesco Mazzoli <f@mazzo.li>"]
 
 repository = "https://github.com/TyOverby/bincode"
@@ -14,7 +14,7 @@ description = "A binary serialization / deserialization strategy and implementat
 rustc-serialize = "0.3.*"
 byteorder = "0.4.*"
 num = "0.1.*"
-serde = "0.6.*"
+serde = "0.7.*"
 
 [dev-dependencies]
-serde_macros = "0.6.*"
+serde_macros = "0.7.*"
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -9,6 +9,7 @@ use byteorder::{BigEndian, ReadBytesExt};
 use num;
 use serde_crate as serde;
 use serde_crate::de::value::ValueDeserializer;
+use serde_crate::de::Deserializer as SerdeDeserializer;
 
 use ::SizeLimit;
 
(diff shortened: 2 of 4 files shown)


nox:
Since power comes with responsibility, need a review? :)

TyOverby:
@nox That would be great!  Also, do you know when servo plans on moving to serde-0.7?

nox:
There is a Webrender workweek ongoing so we are waiting until it finishes to not make their work harder. So next week probably, when we will do a rustup.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bincode"
-version = "0.4.0"
+version = "0.4.1"
```
nox: I think it should be 0.5.0.

src/serde/reader.rs
```diff
@@ -46,10 +47,8 @@ pub enum DeserializeError {
     /// If decoding a message takes more than the provided size limit, this
     /// error is returned.
     SizeLimit,
-    SyntaxError(String),
-    EndOfStreamError,
-    UnknownFieldError,
-    MissingFieldError,
+    EndOfStream,
```
nox: `serde::de::value::Error` has a `EndOfStreamError`, I don't think there is any need to duplicate it.

TyOverby: That's true, but outside of serde, you have `ByteOrderError::UnexpectedEndOfStream`.  I thought it would be weird to coerce a ByteOrderError into a `serde::de::value::Error`, but I could be convinced otherwise.

Cargo.toml
TyOverby: Yep.  I don't know how this sneaked in.  `0.4.1` was the release that fixed the dependencies.  This PR is certainly a breaking change.

src/serde/reader.rs
nox: We can very well special case `DeserializeError::Serde(de::value::Error::EndOfStreamError)`, no? I find it weird to have two different EndOfStream errors.

TyOverby: Yeah, that's reasonable.

src/serde/reader.rs
```diff
@@ -506,7 +481,8 @@ impl<'a, R: Read> serde::de::VariantVisitor for Deserializer<'a, R> {
     {
         let index: u32 = try!(serde::Deserialize::deserialize(self));
         let mut deserializer = (index as usize).into_deserializer();
-        Ok(try!(serde::Deserialize::deserialize(&mut deserializer)))
+        let attempt: Result<V, serde::de::value::Error> = serde::Deserialize::deserialize(&mut deserializer);
+        Ok(try!(attempt))
```
nox: I guess this is needed for type inference to be happy? :(

nox:
Apart from these few remarks, LGTM!

src/serde/reader.rs
TyOverby: yep :(

nox: You should be able to do:

```rust
Ok(try!(serde::Deserialize::deserialize::<Result<V, _>>(&mut deserializer)))
```

I'm not sure it's an improvement though.

TyOverby:
As soon as the lights go green I'll merge!

Issue #41, opened 2015-05-28
decoding malformed Vec<> can cause unrecoverable OOMs irregardless of SizeLimit


jmesmon:
Example:
```
extern crate rustc_serialize;
extern crate bincode;

use std::io::Cursor;

#[derive(RustcEncodable)]
struct FakeVec {
    len: u64,
    byte: u8
}

fn main() {

    let x = bincode::encode(&FakeVec { len: 0xffffffffffffffffu64, byte: 1 }, bincode::SizeLimit::Bounded(10)).unwrap();
    let y : Vec<u8> = bincode::decode_from(&mut Cursor::new(&x[..]), bincode::SizeLimit::Bounded(10)).unwrap();
    println!("{:?} {:?}", y, x);
}
```

Output:
```
oom-bincode % ./target/debug/oom-bincode
zsh: illegal hardware instruction (core dumped)  ./target/debug/oom-bincode
```

gdb backtrace:
```
Core was generated by `./target/debug/oom-bincode'.
Program terminated with signal SIGILL, Illegal instruction.
#0  0x00007f0712686ad0 in oom::hfcd4a1f5b5903c6dX5a ()
warning: Missing auto-load scripts referenced in section .debug_gdb_scripts
of file /home/cody/oom-bincode/target/debug/oom-bincode
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) bt
#0  0x00007f0712686ad0 in oom::hfcd4a1f5b5903c6dX5a ()
#1  0x00007f071266c2a4 in oom_bincode::vec::Vec<T>::with_capacity (capacity=18446744073709551615)
    at /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcollections/vec.rs:223
#2  0x00007f0712670ca4 in fnfn (d=0x7ffd68fec418, len=18446744073709551615)
    at /home/cody/.multirust/toolchains/nightly/cargo/registry/src/github.com-1ecc6299db9ec823/rustc-serialize-0.3.14/src/serialize.rs:458
#3  0x00007f071266f40c in oom_bincode::reader::DecoderReader<'a, R>.Decoder::read_seq<collections::vec::Vec<u8>,closure> (self=0x7ffd68fec418, 
[...]

PR #63, opened 2016-03-10
Respect size limits in read_seq and read_map.

canndrew:
It's possible to cause the bincode rustc_serialise decoder to panic with OOM even if a size limit is set. This PR makes the `read_seq` and `read_map` decoder methods check that the number of elements about to be read is not greater than the number of bytes remaining before we hit the limit. This is not a perfect solution as it sort-of assumes that all types have size 1. For instance, if the size limit is 1KB it should still be allowable to decode a `Vec` of a million `()`s whereas it shouldn't be allowable to decode a `Vec` of 500 `u32`s. With this fix, neither of these situations are handled correctly. However I don't see a solution to this, and having this fix is better than not having it.

This PR also uses `checked_add` in `read_bytes` to so that people can't use malicious data to cause an overflow exception.

### DIFF
--- a/src/rustc_serialize/reader.rs
+++ b/src/rustc_serialize/reader.rs
@@ -347,8 +347,15 @@ impl<'a, R: Read> Decoder for DecoderReader<'a, R> {
         let len = try!(self.read_usize());
         match self.size_limit {
             SizeLimit::Infinite => (),
-            SizeLimit::Bounded(x) if self.read.saturating_add(len as u64) <= x => (),
-            SizeLimit::Bounded(_) => return Err(DecodingError::SizeLimit),
+            SizeLimit::Bounded(x) => {
+                let overflow = match self.read.checked_add(len as u64) {
+                    Some(y) => y > x,
+                    None => true,
+                };
+                if overflow {
+                    return Err(DecodingError::SizeLimit);
+                }
+            },
         };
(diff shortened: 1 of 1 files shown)


canndrew:
Ah, I just noticed there's been an issue open for this since last May: https://github.com/TyOverby/bincode/issues/41

canndrew:
An alternative way to fix this would be to add a `max_container_size` argument to `decode_from`.

nox:
Please squash the two commits and I'll merge this. Thanks for your contribution!

nox:
Oh wait sorry never mind. That's why I shouldn't read my mails why I'm leaving home… Misread your commit message. I'll read tonight what's the story about #41 and try to find a better solution.

canndrew:
@TyOverby: Would you consider merging this PR for now and reverting it if/when rustc_serialise gets fixed? It's unlikely to break any code, unless there's someone out there serialising big `Vec<()>`s.

TyOverby:
Yep. Merged.

canndrew:
Awesome, thanks for that. Could you bump the version number and publish it so I can depend on it?

TyOverby:
@canndrew Done.  0.5.1 is good to go!

TyOverby:
@canndrew Thanks for your contribution!

PR #71, opened 2016-04-15
Don't pull `rustc-serialize` from `num`

alexbool:
This is a follow-up to the previous PR. I have overlooked that the `num` crate pulls `rustc-serialize` unconditionally (and it isn't used by the way), so I disabled it.


TyOverby:
LGTM

PR #83, opened 2016-07-14
WIP: adjust for new seq/map serialization

oli-obk:
just leaving this here so noone repeats any work when doing the next serde update


TyOverby:
:+1:

PR #84, opened 2016-07-28
Update to serde 0.8.0

dtolnay:
Supersedes #83.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,10 @@ description = "A binary serialization / deserialization strategy and implementat
 rustc-serialize = { version = "0.3.*", optional = true }
 byteorder = "0.5.*"
 num-traits = "0.1.32"
-serde = { version = "0.7.*", optional = true }
+serde = { version = "0.8.*", optional = true }
 
 [dev-dependencies]
-serde_macros = "0.7.*"
+serde_macros = "0.8.*"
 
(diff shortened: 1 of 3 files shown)


TyOverby:
Does this serde update include any changes that would change the ordering of bytes produced by bincode?

dtolnay:
No. There are 3 user-facing changes but all are in deserialization:

- `deserialize_str` used to not be implemented, meaning it forwarded to `deserialize` and failed. Now it is implemented to deserialize a `&str`. This affects only Deserialize types that used to fail and now succeed.
- `deserialize_struct_field` fails with a better message.
- `deserialize_ignored_any` fails with a better message.

As an incentive, I suspect you will see a nice performance improvement in serialization for both structs and vectors.

I can put together a test suite this weekend before you release to prove the bytes are the same.

TyOverby:
Sounds good.  I was only asking because I wanted to know if I'd need to bump the version number.  However, I'm pretty sure I need to bump the version anyway, simply because it's re-exporting serde which bumped to 0.8.0.

The test suite already makes sure that the `rustc-serialize` encoding is the same (byte for byte) as the `serde` encoding, so I'm not super worried.

Thanks for the contribution!

TyOverby:
Published to crates.io as 0.6.0

PR #89, opened 2016-11-23
Delegate emit_enum_struct_variant calls to emit_enum_variant

nbaksalyar:
[`read_enum_struct_variant`](https://github.com/TyOverby/bincode/blob/f06a4cfcb5b194e54d4997c200c75b88b6c3fba4/src/rustc_serialize/reader.rs#L279-L283) delegates calls to [`read_enum_variant`](https://github.com/TyOverby/bincode/blob/f06a4cfcb5b194e54d4997c200c75b88b6c3fba4/src/rustc_serialize/reader.rs#L260-L263) which expects `u32` prefix to encode a variant number. Previous implementation of [`emit_enum_struct_variant`](https://github.com/TyOverby/bincode/blob/f06a4cfcb5b194e54d4997c200c75b88b6c3fba4/src/rustc_serialize/writer.rs#L178-L187) doesn't emit it, making serialized data invalid.

### DIFF
--- a/src/rustc_serialize/writer.rs
+++ b/src/rustc_serialize/writer.rs
@@ -176,14 +176,14 @@ impl<'a, W: Write> Encoder for EncoderWriter<'a, W> {
         f(self)
     }
     fn emit_enum_struct_variant<F>(&mut self,
-                                   _: &str,
-                                   _: usize,
-                                   _: usize,
+                                   v_name: &str,
+                                   v_id: usize,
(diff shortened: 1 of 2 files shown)


TyOverby:
Looks good, thanks for catching this!  Could you add a regression test that would fail on this case?

nbaksalyar:
Does it look better? :)

TyOverby:
Looks good, thanks!

PR #90, opened 2016-12-12
More efficient serialize_bytes

dtolnay:
Reported in IRC: user blank_name2 tried serializing `HashMap<String, HashSet<Bytes>>` vs `HashMap<String, HashSet<&Path>>`, the `&Path` version was done in ~.6 seconds while the `&[u8]` one took a full 3 seconds more.

This PR resolves the difference between the two while keeping the same serialized form.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ version = "0.8.*"
 optional = true
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


TyOverby:
Looks good to me!  Thanks for the contribution!

Issue #86, opened 2016-08-27
fixed length arrays have their size encoded


oohal:
When a encoding a structure that contains a fixed length array the length of the array is also encoded. For example, this code:

``` rust
extern crate rustc_serialize;
extern crate bincode;

use bincode::SizeLimit;
use bincode::rustc_serialize::encode;

#[derive(RustcEncodable)]
struct TestStruct {
    a : u16,
    b : [u8; 6],
}

fn main() {
    let test = TestStruct {a: 0xffff, b: [1, 2, 3, 4, 5, 6]};
    let encoded: Vec<u8> = encode(&test, SizeLimit::Infinite).unwrap();

    println!("header: {:?}", encoded);
}
```

Produces this output:

```
header: [255, 255, 0, 0, 0, 0, 0, 0, 0, 6, 1, 2, 3, 4, 5, 6]
```

This behaviour makes it hard to use bincode for unpacking externally defined binary formats (e.g ELF headers). I suppose this library isn't really designed for that, but it works well enough for BE formats so it would be nice to see this fixed in the next major version.

TyOverby:
I don't think that there is a way to do this with rustc-serialize.  At least, there isn't a trait method for statically sized arrays.

Matvey-Kuk:
+1 for this idea

TyOverby:
This feature is only possible if we drop RustcSerialize support.  I'm planning on doing that eventually, but it's on the backlist until serde hits 1.0

Matvey-Kuk:
@TyOverby thank you! Just wanted to mention there is somebody else who need this =)

aldanor:
Very much in support of this as well. Now that serde 0.9 is out, is it a good time to revisit this?

darfink:
I'd like to chime in on this as well. Working with external data and this issue hit me, ended up having to use a fork.

mikedilger:
I also ran into this.  What if we change the code to skip these lengths if the `rustc_serialize` feature is disabled?

TyOverby:
Since there's so much interest in this, I think I'll begin work on removing rustc-serialize support from bincode altogether.  I've wanted to do this for a long time, but since serde is hitting 1.0 and macros are landing, I think it'll be pretty reasonable to move on to.

TyOverby:
rustc-serialize removal happening here: https://github.com/TyOverby/bincode/pull/95

aldanor:
[...]

PR #93, opened 2017-01-31
Update to serde 0.9.X

### LINKED ISSUE #92: Support serde 0.9.X
Serde 0.9.0-rc1 is out (https://github.com/serde-rs/serde/issues/668) and there are some breaking changes.
It would be great to have a support for it. Thanks.

dgriffen:
Closes #92 

One of the few questions I had left after doing this was what to do with the `isize_invalid_deserialize` test. The test used variants from the `serde::de::value::Error` enum. This type is now a struct and doesn't have variants anymore so the test wouldn't work without string parsing, which seemed fragile.

### DIFF
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -54,7 +54,7 @@ impl Error for DeserializeError {
             DeserializeError::IoError(ref err) => Error::description(err),
             DeserializeError::InvalidEncoding(ref ib) => ib.desc,
             DeserializeError::SizeLimit => "the size limit for decoding has been reached",
-            DeserializeError::Custom(_) => "a custom serialization error was reported",
+            DeserializeError::Custom(ref msg) => msg,
 
         }
     }
@@ -75,12 +75,6 @@ impl From<IoError> for DeserializeError {
     }
 }
 
-impl From<serde::de::value::Error> for DeserializeError {
-    fn from(err: serde::de::value::Error) -> DeserializeError {
-        DeserializeError::Custom(err.description().to_string())
-    }
-}
-
 impl fmt::Display for DeserializeError {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         match *self {
@@ -91,7 +85,7 @@ impl fmt::Display for DeserializeError {
             DeserializeError::SizeLimit =>
                 write!(fmt, "SizeLimit"),
             DeserializeError::Custom(ref s) =>
-                write!(fmt, "Custom Error {}", s),
+                s.fmt(fmt),
         }
     }
 }
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -97,7 +97,7 @@ impl fmt::Display for DeserializeError {
 }
 
 impl serde::de::Error for DeserializeError {
-    fn custom<T: ::std::fmt::Display>(desc: T) -> DeserializeError {
+    fn custom<T: fmt::Display>(desc: T) -> DeserializeError {
         DeserializeError::Serde(serde::de::value::Error::custom(desc))
     }
 }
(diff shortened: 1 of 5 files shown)


src/serde/reader.rs
```diff
@@ -303,162 +275,193 @@ impl<'a, R: Read> serde::Deserializer for Deserializer<'a, R> {
...
+            fn visit_variant_seed<V>(self, seed: V) -> DeserializeResult<(V::Value, Self::Variant)>
+                where V: serde::de::DeserializeSeed,
+            {
+                let idx: u32 = try!(serde::de::Deserialize::deserialize(&mut *self.0));
+                let val: Result<_, DeserializeError> = seed.deserialize(idx.into_deserializer());
+                Ok((try!(val), self))
+            }
+        }
+
+        impl<'a, R: Read + 'a> serde::de::VariantVisitor for VariantVisitor<'a, R> {
```
dtolnay: We already have a VariantVisitor below don't we? Let's use that one.

```rust
impl<'a, R: Read> serde::de::VariantVisitor for &'a mut Deserializer<R> { /* ... */ }
```

src/serde/reader.rs
```diff
@@ -303,162 +275,193 @@ impl<'a, R: Read> serde::Deserializer for Deserializer<'a, R> {
...
+    fn deserialize_enum<V>(self,
                      _enum: &'static str,
                      _variants: &'static [&'static str],
                      mut visitor: V) -> Result<V::Value, Self::Error>
-        where V: serde::de::EnumVisitor,
+        where V: serde::de::Visitor,
     {
-        visitor.visit(self)
-    }
+        struct VariantVisitor<'a, R: Read + 'a>(&'a mut Deserializer<R>);
```
dtolnay: I think this would be clearer without the wrapper type:

```rust
impl<'a, R: Read> serde::de::EnumVisitor for &'a mut Deserializer<R>
```

src/serde/reader.rs
```diff
@@ -97,12 +97,8 @@ impl fmt::Display for DeserializeError {
 }
 
 impl serde::de::Error for DeserializeError {
-    fn custom<T: Into<String>>(desc: T) -> DeserializeError {
-        DeserializeError::Serde(serde::de::value::Error::Custom(desc.into()))
-    }
-
-    fn end_of_stream() -> DeserializeError {
-        DeserializeError::Serde(serde::de::value::Error::EndOfStream)
+    fn custom<T: ::std::fmt::Display>(desc: T) -> DeserializeError {
```
dtolnay: The `fmt` module is already imported above so this could be `fmt::Display` instead of `::std::fmt::Display`.

src/serde/writer.rs
```diff
@@ -29,17 +29,17 @@ pub enum SerializeError {
...
 }
 
 fn wrap_io(err: IoError) -> SerializeError {
     SerializeError::IoError(err)
 }
 
 impl serde::ser::Error for SerializeError {
-    fn custom<T: Into<String>>(msg: T) -> Self {
-        SerializeError::Custom(msg.into())
+    fn custom<T: ::std::fmt::Display>(msg: T) -> Self {
```
dtolnay: Please use `fmt::Display` instead of `::std::fmt::Display`.

src/serde/writer.rs
```diff
@@ -29,17 +29,17 @@ pub enum SerializeError {
...
+        SerializeError::Custom(format!("{}", msg))
```
dtolnay: I think this would be clearer as `msg.to_string()` instead of `format!("{}", msg)`.

dgriffen: Should I make these updates as separate commits? or squash all the changes into one commit?

dtolnay:
Regarding the error tests - I am not sure why `DeserializeError::Serde` contains `serde::de::value::Error`. I think it should contain String instead, just like the `SerializeError`. And the tests should do string matching - this is what serde_json does.

src/serde/writer.rs
dtolnay: Please make any changes as new commits. GitHub's code review has very poor support for reviewing squashed / amended / rebased commits so I would have to review the whole thing all over again to see what you changed. If he prefers, @TyOverby can squash or rebase the whole thing at the end when the review is done and it is ready to merge.

dgriffen:
Should I re-add the test and change `DeserializeError::Serde` to contain a string then?

dtolnay:
Yes please change the error to a String and update the deserializing_errors test to check the error messages.

src/serde/reader.rs
```diff
@@ -54,7 +54,7 @@ impl Error for DeserializeError {
             DeserializeError::IoError(ref err) => Error::description(err),
             DeserializeError::InvalidEncoding(ref ib) => ib.desc,
             DeserializeError::SizeLimit => "the size limit for decoding has been reached",
-            DeserializeError::Serde(ref s) => s.description(),
+            DeserializeError::Custom(_) => "a custom serialization error was reported",
```
dtolnay: You should be able to return the real error here. `DeserializeError::Custom(ref msg) => msg`

src/serde/reader.rs
```diff
@@ -77,7 +77,7 @@ impl From<IoError> for DeserializeError {
 
 impl From<serde::de::value::Error> for DeserializeError {
```
dtolnay: We don't need this impl anymore.

src/serde/reader.rs
```diff
@@ -90,19 +90,15 @@ impl fmt::Display for DeserializeError {
                 write!(fmt, "InvalidEncoding: {}", ib),
             DeserializeError::SizeLimit =>
                 write!(fmt, "SizeLimit"),
-            DeserializeError::Serde(ref s) =>
-                s.fmt(fmt),
+            DeserializeError::Custom(ref s) =>
+                write!(fmt, "Custom Error {}", s),
```
dtolnay: I would prefer to just format the message like before - `s.fmt(fmt)`. I don't think "Custom Error" adds any clarity or context.

dgriffen: Sounds good. I was just following the convention used by SerializeError

dtolnay: A major focus of serde 0.9 is better error message so as the person doing this upgrade, part of your job is improving existing bad error messages!

dtolnay:
I tried compiling this and I get a ton of unused_import, unused_mut, and unused_variables warnings. The master branch builds cleanly so please fix all of the warnings.

dgriffen:
I wasn't aware those errors were introduced with my changes. I'll get those corrected.

TyOverby:
This is really great work @dgriffen, thanks for taking this PR on!

From the bincode API side, this all looks good to me.  @dtolnay: is there anything remaining as far as integration with serde is concerned?

TyOverby:
🎈 🎈

dgriffen:
@TyOverby are there any more blocking issues that a new release is waiting on? I'd be willing to work on them.

dtolnay:
Possibly https://github.com/TyOverby/bincode/issues/86.

TyOverby:
@dgriffen: #86 and #94 were the next ones that I was going to take on. 

I also just opened up #95 which is a precondition for #86.

TyOverby:
@dgriffen: I also just published `1.0.0-alpha1` which includes your serde 0.9.* changes *and* my rustc-serialize removal.

PR #97, opened 2017-01-31
change error message

TyOverby:
I didn't like the old error message

### DIFF
--- a/src/serde/writer.rs
+++ b/src/serde/writer.rs
@@ -5,7 +5,6 @@ use std::io::Write;
 use std::u32;
 
 use serde_crate as serde;
-use serde_crate::ser::Error as SeError;
 
 use byteorder::{BigEndian, WriteBytesExt};
 
@@ -22,6 +21,7 @@ pub enum SerializeError {
     /// This error is returned before any bytes are written to the
     /// output `Writer`.
     SizeLimit,
+    InfiniteSequence,
     /// A custom error message
     Custom(String)
 }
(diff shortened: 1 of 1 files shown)


TyOverby:
@dtolnay

dtolnay:
Seems reasonable. Another option would be to add a SerializeError variant for this case.

TyOverby:
I like that more actually.

dtolnay:
I don't think "infinite" is the right word, especially because SizeLimit::Infinite means something completely different. If serialize_seq does not receive a size, it is because the Serialize type does not know the length up front, for example because we are serializing an iterator that can only be iterated once.

src/serde/writer.rs
```diff
@@ -49,6 +49,7 @@ impl fmt::Display for SerializeError {
         match *self {
             SerializeError::IoError(ref err) => write!(f, "IoError: {}", err),
             SerializeError::Custom(ref s) => write!(f, "Custom Error {}", s),
+            SerializeError::InfiniteSequence => write!(f, "InfiniteSequence"),
```
dtolnay: Please make this a human-readable error message. This is the Display impl not Debug.

TyOverby:
@dtolnay Changed it to `SequenceMustHaveLength` and the Display impl mentions maps and seqs

dtolnay:
Cool. I filed https://github.com/TyOverby/bincode/issues/98 to follow up on the other messages.

PR #101, opened 2017-01-31
Remove sized bounds where appropriate

### LINKED ISSUE #99: Accept dynamically sized Read, Write, and Serialize implementations
See https://github.com/TyOverby/bincode/pull/95#issuecomment-276464744.

Sure! Accepting a ?Sized type parameter for io::Read and io::Write means that it will work with &mut io::Read and &mut io::Write trait objects which are dynamically sized.

```rust
use std::fs::File;
use std::io::Read;

fn main() {
    let mut data = File::open("important.bin").unwrap();
    let trait_object: &mut Read = &mut data;
    deserialize_from(trait_object); // requires ?Sized
}

fn deserialize_from<R>(_: &mut R) where R: Read {}
```
And accepting a ?Sized type parameter for &T means that it can work with dynamically sized implementations of Serialize, which includes str and [T] among others.

```rust
fn main() {
    serialize("&str"); // requires ?Sized
}

trait Serialize {}
impl Serialize for str {}

fn serialize<T>(_: &T) where T: Serialize {}
```

dgriffen:
Closes #99

### DIFF
--- a/src/serde/mod.rs
+++ b/src/serde/mod.rs
@@ -34,7 +34,8 @@ mod writer;
 /// writer is in an invalid state, as writing could bail out in the middle of
 /// serializing.
 pub fn serialize_into<W, T>(writer: &mut W, value: &T, size_limit: SizeLimit) -> SerializeResult<()>
-    where W: Write, T: serde::Serialize,
+    where W: Write + ?Sized, 
+          T: serde::Serialize + ?Sized,
 {
     match size_limit {
         SizeLimit::Infinite => { }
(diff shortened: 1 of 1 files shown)


TyOverby:
On the stable compiler, you still need to have the `: ?Sized` inside of the `<..>` for the function (it can't be in the `where` clause)

TyOverby:
Also, I'd like to point out that you beat me to this PR by 3 minutes 
💯 :100:

dgriffen:
I'm at dinner right now. I'll fix after.

TyOverby:
Looks good!

PR #102, opened 2017-02-01
Remove length encoding

### LINKED ISSUE #86: fixed length arrays have their size encoded

dgriffen:
Closes #86

### DIFF
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -398,6 +398,6 @@ fn bytes() {
     let data = b"abc\0123";
     let s = serialize(&data, Infinite).unwrap();
     let s2 = serialize(&Bytes::new(data), Infinite).unwrap();
-    assert_ne!(s, s2);
+    assert_eq!(s[..], s2[8..]);
 }
 
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -398,6 +398,6 @@ fn bytes() {
     let data = b"abc\0123";
     let s = serialize(&data, Infinite).unwrap();
     let s2 = serialize(&Bytes::new(data), Infinite).unwrap();
-    assert_eq!(s, s2);
+    assert_ne!(s, s2);
 }
 
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -336,28 +336,8 @@ impl<'a, R: Read> serde::Deserializer for &'a mut Deserializer<R> {
     }
 
     fn deserialize_seq_fixed_size<V>(self,
-                            _: usize,
+                            len: usize,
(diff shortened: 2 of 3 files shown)


tests/test.rs
```diff
@@ -398,6 +398,6 @@ fn bytes() {
     let data = b"abc\0123";
     let s = serialize(&data, Infinite).unwrap();
     let s2 = serialize(&Bytes::new(data), Infinite).unwrap();
-    assert_eq!(s, s2);
+    assert_ne!(s, s2);
```
dgriffen: So I'm looking at the tests we already have. We do fixed sized array tests so just doing `assert_eq!(s, vec![ byte values here ])` wouldn't really add anything to the test coverage. What I could do is `assert_eq!(s[..], s2[8..]);` to verify that after the length encoding they are identical.

tests/test.rs
```diff
@@ -398,6 +398,6 @@ fn bytes() {
...
 }
```
dgriffen: Since we do tests for fixed sized arrays, is adding a test for both serialization and deserialization really necessary?

tests/test.rs
TyOverby: > assert_eq!(s[..], s2[8..]); to verify that after the length encoding they are identical.

That sounds good; I like it!

tests/test.rs
TyOverby: Looking at this some more, I think that my request is out-of-scope for this PR.  However, the structures serde::bytes::Bytes and serde::bytes::ByteBuf have *special handling* that we certainly want to be testing.  I'll file another issue for this.

PR #103, opened 2017-02-01
Make Reader and Writer generic on Endianness

TyOverby:


src/serde/mod.rs
```diff
@@ -69,7 +70,7 @@ pub fn serialize<T>(value: &T, size_limit: SizeLimit) -> SerializeResult<Vec<u8>
         SizeLimit::Infinite => Vec::new()
     };
 
-    try!(serialize_into(&mut writer, value, SizeLimit::Infinite));
+    try!(serialize_into::<_, _, ::byteorder::BigEndian>(&mut writer, value, SizeLimit::Infinite));
```
aldanor: `BigEndian` here?

src/serde/mod.rs
```diff
@@ -107,7 +108,7 @@ pub fn deserialize_from<R, T>(reader: &mut R, size_limit: SizeLimit) -> Deserial
     where R: Read,
           T: serde::Deserialize,
 {
-    let mut deserializer = Deserializer::new(reader, size_limit);
+    let mut deserializer = Deserializer::<_, BigEndian>::new(reader, size_limit);
```
aldanor: `deserialize_from` not generic over byte order?

TyOverby:
@aldanor: in its current state, this PR adds endian genericness to the backend, but not the public functions.  That will change in the future, but I wanted to get the difficult part passing tests first.

aldanor:
@TyOverby I see then, thanks; looks good to me given that this will be later exposed to the crate-level functions.

// Should there be some explicit tests for big/little endian?

TyOverby:
@aldanor: I've gone with the decision to add an `endian_choice` module with endianness exposed.

TyOverby:
Oh, also, I've picked LittleEndian as the new default endianness (mainly for speed)

tests/test.rs
```diff
@@ -398,3 +398,11 @@ fn bytes() {
     assert_eq!(s[..], s2[8..]);
 }
 
+
+#[test]
+fn endian_difference() {
+    let x = 10u64;
+    let little = serialize::<_, byteorder::LittleEndian>(&x, Infinite).unwrap();
+    let big = serialize::<_, byteorder::BigEndian>(&x, Infinite).unwrap();
+    assert!(little != big);
```
dgriffen: Wouldn't it be better to use `assert_ne!(little, big);`? As I understand it you get better error messages that way.

TyOverby: huh, I didn't actually know about that one

aldanor:
That's one way to go about it I guess. Little endian as a default sounds grand; worth mentioning in the docstrings of the root module somewhere?

Issue #98, opened 2017-01-31
Polish all error messages


dtolnay:
The current error messages have an unfriendly Display representation. For example SerializeError in https://github.com/TyOverby/bincode/commit/9e47b71af6476c81ec3036c900399ac903b5fc40:

```rust
impl fmt::Display for SerializeError {
    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
        match *self {
            SerializeError::IoError(ref err) => write!(f, "IoError: {}", err),
            SerializeError::Custom(ref s) => write!(f, "Custom Error {}", s),
            SerializeError::InfiniteSequence => write!(f, "InfiniteSequence"),
            SerializeError::SizeLimit => write!(f, "SizeLimit"),
        }
    }
}
```

Let's audit all the error messages in this crate and make sure everything has a self-explanatory message that a human can read and understand what went wrong.

TyOverby:
I think a neat feature of the errors (once merged) would be tracking if they came from encoding or decoding. 

Example: 

`While decoding, bincode hit an IoError from the underlying reader: $IOERRORTEXT`
vs
`While encoding, bincode hit an IoError from the underlying writer: $IOERRORTEXT`

dtolnay:
I filed individual issues for the error message problems I noticed.

TyOverby:
Closed all of these (I think)

PR #104, opened 2017-02-01
Combine error types

### LINKED ISSUE #100: Combine SerializeError and DeserializeError, SerializeResult and DeserializeResult
See https://github.com/TyOverby/bincode/pull/95#issuecomment-276457709.

dgriffen:
Closes #100

### DIFF
--- a/src/serde/mod.rs
+++ b/src/serde/mod.rs
@@ -127,7 +127,7 @@ impl fmt::Display for InvalidEncoding {
 /// If this returns an `Error` (other than SizeLimit), assume that the
 /// writer is in an invalid state, as writing could bail out in the middle of
 /// serializing.
-pub fn serialize_into<W: ?Sized, T: ?Sized>(writer: &mut W, value: &T, size_limit: SizeLimit) -> SerializeResult<()>
+pub fn serialize_into<W: ?Sized, T: ?Sized>(writer: &mut W, value: &T, size_limit: SizeLimit) -> Result<()>
     where W: Write, 
           T: serde::Serialize,
 {
@@ -147,7 +147,7 @@ pub fn serialize_into<W: ?Sized, T: ?Sized>(writer: &mut W, value: &T, size_limi
 ///
 /// If the serialization would take more bytes than allowed by `size_limit`,
 /// an error is returned.
-pub fn serialize<T: ?Sized>(value: &T, size_limit: SizeLimit) -> SerializeResult<Vec<u8>>
+pub fn serialize<T: ?Sized>(value: &T, size_limit: SizeLimit) -> Result<Vec<u8>>
     where T: serde::Serialize
 {
     // Since we are putting values directly into a vector, we can do size
@@ -202,7 +202,7 @@ pub fn serialized_size_bounded<T: ?Sized>(value: &T, max: u64) -> Option<u64>
 /// If this returns an `Error`, assume that the buffer that you passed
 /// in is in an invalid state, as the error could be returned during any point
 /// in the reading.
-pub fn deserialize_from<R: ?Sized, T>(reader: &mut R, size_limit: SizeLimit) -> DeserializeResult<T>
+pub fn deserialize_from<R: ?Sized, T>(reader: &mut R, size_limit: SizeLimit) -> Result<T>
     where R: Read,
           T: serde::Deserialize,
 {
--- a/src/serde/mod.rs
+++ b/src/serde/mod.rs
@@ -3,19 +3,16 @@
 //! implementation.
 
 use std::io::{Write, Read};
+use std::io::Error as IoError;
+use std::{error, fmt, result};
 use ::SizeLimit;
 
 pub use self::reader::{
     Deserializer,
-    DeserializeResult,
-    DeserializeError,
-    InvalidEncoding
 };
 
 pub use self::writer::{
     Serializer,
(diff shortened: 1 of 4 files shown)


dtolnay:
I can recommend what I have done for `serde::de::value::Error`, `serde_json::Error` and `serde_yaml::Error` which is hide them behind an opaque newtype wrapper.

```rust
pub struct Error(Box<ErrorImpl>);

enum ErrorImpl {
    /* variants */
}
```

The Box is to drop the size of Error from 48 bytes down to 8 bytes which sped up performance of all of serde_json by 5%, and the newtype wrapper is to let us add or change error variants without a breaking change. Nobody actually pattern-matches on these errors - they just print the message and fail.

dgriffen:
Sounds like a good idea. @TyOverby what do you think?

TyOverby:
@dtolnay @dgriffen: I'm all for sticking it inside of a Box, but my use-cases for bincode require pattern matching on the error: I *really* care if the deserialization failed because of a dropped network connection (IoError) or because they sent bad bytes over the wire (programmer error).

TyOverby:
I am surprised that allocating is cheaper than moving an extra 40 bytes though.

TyOverby:
Oh, I guess if you rarely hit error cases, then it's a pretty good speedup in the general case.

dgriffen:
Oops, looks like I made a mistake when fixing conflicts. That's the last time I use the Github's conflict resolver.

dgriffen:
So how should I move forward with any changes to `Error`? It sounds like we either change the `Result` type to carry a `Box<Error>` or do what @dtolnay said, but make `ErrorImpl` public and give it a better name if we want pattern matching.

TyOverby:
@dgriffen: yeah, that sounds like a good move forward!  Off the top of my head I can't think of a good enum name though.  `ErrorKind` maybe?

dgriffen:
`ErrorKind` sounds good to me.

TyOverby:
I tried to resolve the merge conflicts with githubs online tool.  Let's see if it actually worked!

dgriffen:
Looks like you kept the old error types

TyOverby:
Well, that's the last time I use that tool.

dgriffen:
https://github.com/TyOverby/bincode/pull/104#issuecomment-276570084

TyOverby:
And here I thought I was helping :|

dgriffen:
Looks like a few more merge errors that didnt get caught, I'll clean those up

TyOverby:
@dtolnay Want to do another pass?

src/serde/mod.rs
```diff
@@ -25,15 +22,115 @@ use serde_crate as serde;
...
+    SequenceMustHaveLength,
+    Custom(String)
+}
+
+impl error::Error for ErrorKind {
+    fn description(&self) -> &str {
+        match *self {
+            ErrorKind::IoError(ref err) => error::Error::description(err),
+            ErrorKind::InvalidEncoding(ref ib) => ib.desc,
+            ErrorKind::SequenceMustHaveLength => "bincode can't encode infinite sequences",
```
dtolnay: I don't think this error is accurate - the one you have in Display is a better characterization of SequenceMustHaveLength.

src/serde/mod.rs
```diff
@@ -25,15 +22,115 @@ use serde_crate as serde;
...
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ErrorKind::IoError(ref ioerr) =>
+                write!(fmt, "IoError: {}", ioerr),
+            ErrorKind::InvalidEncoding(ref ib) =>
+                write!(fmt, "InvalidEncoding: {}", ib),
+            ErrorKind::SequenceMustHaveLength =>
+                write!(fmt, "Bincode can only encode sequences and maps that have a knowable size ahead of time."),
+            ErrorKind::SizeLimit =>
+                write!(fmt, "SizeLimit"),
```
dtolnay: This is the Display representation not the Debug representation, so please make this message self-explanatory.

src/serde/mod.rs
```diff
@@ -25,15 +22,115 @@ use serde_crate as serde;
...
+    fn from(err: IoError) -> Error {
+        ErrorKind::IoError(err).into()
+    }
+}
+
+impl fmt::Display for ErrorKind {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ErrorKind::IoError(ref ioerr) =>
+                write!(fmt, "IoError: {}", ioerr),
```
dtolnay: Rust identifiers like `IoError` belong in the Debug representation. For the Display representation I would prefer "IO error: {}".

src/serde/mod.rs
```diff
@@ -25,15 +22,115 @@ use serde_crate as serde;
...
+            ErrorKind::InvalidEncoding(ref ib) =>
+                write!(fmt, "InvalidEncoding: {}", ib),
```
dtolnay: Ditto here - "invalid encoding: {}".

dtolnay:
The code looks good but the error messages are not as helpful as they could be. Ok to merge and follow up under #98 in another PR.

TyOverby:
Looks good!  We'll take care of error messages as a part of #98

PR #108, opened 2017-02-03
Run tests on all channels

TyOverby:


TyOverby:
I truly love diffs like this

PR #111, opened 2017-02-10
Remove InvalidEncoding struct

### LINKED ISSUE #110: Get rid of "InvalidEncoding" struct.
The fields of this struct can go right in the Error enum

dgriffen:
Closes #110 

Open questions:  
Do we want to use a struct-like variant or a tuple-like variant? Right now I implemented it as a struct-like.


TyOverby:
Struct-like is what I was imagining

dgriffen:
Well then I guess I made the right choice.

PR #115, opened 2017-02-22
Smarter deserialize_bytes and deserialize_byte_buf

tikue:
It seems we can just read all at once rather than reading byte by byte.

### DIFF
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -55,9 +55,11 @@ impl<R: Read> Deserializer<R> {
         let len = try!(serde::Deserialize::deserialize(&mut *self));
         try!(self.read_bytes(len));
 
-        let mut buffer = Vec::new();
-        try!(self.reader.by_ref().take(len as u64).read_to_end(&mut buffer));
-        Ok(buffer)
+        let len = len as usize;
+        let mut bytes = Vec::with_capacity(len);
+        unsafe { bytes.set_len(len); }
+        try!(self.reader.read_exact(&mut bytes));
+        Ok(bytes)
     }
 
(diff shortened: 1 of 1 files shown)


TyOverby:
This looks really good, thanks!  While you are in the read_vec code, could you make a little change?

```rust
let mut buffer = Vec::new();
try!(self.reader.by_ref().take(len as u64).read_to_end(&mut buffer));
```

could be replaced with
```
let mut bytes = Vec::with_capacity(len);
unsafe { bytes.set_len(len); }
try!(self.reader.read_exact(&mut bytes[..]));
```

and you'd probably see even better performance.

(this implementation was ruthlessly stolen from https://github.com/TyOverby/bincode/pull/116/files)

src/serde/reader.rs
```diff
@@ -192,13 +195,13 @@ impl<'a, R: Read> serde::Deserializer for &'a mut Deserializer<R> {
     fn deserialize_bytes<V>(self, visitor: V) -> Result<V::Value>
         where V: serde::de::Visitor,
     {
-        self.deserialize_seq(visitor)
+        visitor.visit_byte_buf(try!(self.read_vec()))
```
TyOverby: shouldn't this be a call to `visit_bytes`?

tikue: Done.

tikue:
Great suggestion, thanks -- done!

dtolnay:
Still not as smart as it could be. I filed #118 to follow up in another PR.

TyOverby:
@dtolnay: sounds good!  I'll merge this and start in on the recommendations in #118

PR #116, opened 2017-02-23
Use the fast path for deserializing bytes::{Bytes, ByteBuf}.

khuey:
On a workload I have that makes heavy use of ByteBuf this cuts the time to execute by one order of magnitude.

### DIFF
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -46,6 +46,16 @@ impl<R: Read> Deserializer<R> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


TyOverby:
Thanks for the contribution!  However, it looks like https://github.com/TyOverby/bincode/pull/115 beat you to it by several hours.

Because @tikue got there first I'm going to take that PR.  There are a few minor differences between the PRs, and I'll suggest that @tikue use `read_exact` like you did.

khuey:
Ha, so it did.  Thanks!

Issue #123, opened 2017-02-24
Bincode is 2x-4x slower than serde-bench format


dtolnay:
Check out `cargo bench` in https://github.com/serde-rs/bench. On my computer:

```
test bincode_deserialize ... bench:          42 ns/iter (+/- 0)
test bincode_serialize   ... bench:          82 ns/iter (+/- 1)
test serde_deserialize   ... bench:          24 ns/iter (+/- 0)
test serde_serialize     ... bench:          20 ns/iter (+/- 1)
```

The format is byte-for-byte identical (`cargo test` to confirm) so there should be no reason that Bincode would take 75% longer to deserialize and 310% longer to serialize in the SizeLimit::Infinite / &[u8] case.

In addition to all the tickets I filed today, see if you can identify other optimizations that would close this gap.

cc @maciejhirsz

TyOverby:
I would expect that endianness and the current non-trait-based size limit is responsible for most of this.

dtolnay:
Both are using big-endian. I think it's all size limit.

TyOverby:
I made a benchmarking tool here: https://github.com/TyOverby/bincode-bench

I played around with a few optimizations, and the https://github.com/TyOverby/bincode/pull/127 PR is twice as fast as serde-bench due to the "size-find and then pre-alloc" optimization.

Somehow this is actually faster than having an already-pre-allocated array.

```
bincode writer    : 334332ns 
serde-bench (ser) : 421333ns
bincode to vec    : 197910ns
```

Decoding is almost always the same speed.

TyOverby:
@dtolnay: could you run your benchmark suite to confirm?

dtolnay:
I get similar results. Nice work.

TyOverby:
🎈 🎈 🎈 🎈 🎈

PR #124, opened 2017-02-25
make SizeLimit a trait

### LINKED ISSUE #121: Consider making SizeLimit a trait rather than an enum
```diff
-pub struct Deserializer<R> {
+pub struct Deserializer<R, L> {
     reader: R,
-    size_limit: SizeLimit,
+    size_limit: L,
     read: u64
 }
```

This way the Infinite case would not be slowed down by all the branching in Deserializer::read_bytes:

```rust
match self.size_limit {
    SizeLimit::Infinite => Ok(()),
    SizeLimit::Bounded(x) if self.read <= x => Ok(()),
    SizeLimit::Bounded(_) => Err(ErrorKind::SizeLimit.into())
}
```

Something like:

```rust
trait SizeLimit {
    fn check(&self, read: u64) -> Result<()>;
}

struct Infinite;
impl SizeLimit for Infinite {
    #[inline]
    fn check(&self, _read: u64) -> Result<()> {
        Ok(())
    }
}

struct Bounded(u64);
impl SizeLimit for Bounded {
    fn check(&self, read: u64) -> Result<()> {
        if read <= self.0 {
            Ok(())
        } else {
            Err(ErrorKind::SizeLimit.into())
        }
    }
}

impl<R: Read, L: SizeLimit> Deserializer<R, L> {
    fn read_bytes(&mut self, count: u64) -> Result<()> {
        self.read += count;
        self.size_limit.check(self.read)
    }
}
```

The compiler may be able to generate faster code in the Infinite case knowing that no branching is happening.

TyOverby:
Closes #121

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,12 +17,12 @@
 //!
 //! ```rust
 //! extern crate bincode;
-//! use bincode::{serialize, deserialize};
+//! use bincode::{serialize, deserialize, Bounded};
 //! fn main() {
 //!     // The object that we will serialize.
 //!     let target = Some("hello world".to_string());
 //!     // The maximum size of the encoded message.
-//!     let limit = bincode::SizeLimit::Bounded(20);
+//!     let limit = Bounded(20);
 //!
 //!     let encoded: Vec<u8>        = serialize(&target, limit).unwrap();
 //!     let decoded: Option<String> = deserialize(&encoded[..]).unwrap();
@@ -51,7 +51,7 @@ use std::io::{Read, Write};
 
 pub use serde::{ErrorKind, Error, Result, serialized_size, serialized_size_bounded};
 
-pub type Deserializer<W> = serde::Deserializer<W, byteorder::LittleEndian>;
+pub type Deserializer<W, S> = serde::Deserializer<W, S, byteorder::LittleEndian>;
 pub type Serializer<W> = serde::Serializer<W, byteorder::LittleEndian>;
 
 /// Deserializes a slice of bytes into an object.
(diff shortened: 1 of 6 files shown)


TyOverby:
cc @dtolnay

dtolnay:
This didn't make much of a difference in #123...

#### Before

```
test bincode_deserialize ... bench:          42 ns/iter (+/- 0)
test bincode_serialize   ... bench:          82 ns/iter (+/- 1)
```

#### After

```
test bincode_deserialize ... bench:          36 ns/iter (+/- 1)
test bincode_serialize   ... bench:          84 ns/iter (+/- 4)
```

I don't have time to review the code tonight but maybe read through the code again and focus on places where the Infinite / &[u8] case has different behavior compared to my implementation.

src/lib.rs
```diff
@@ -119,8 +118,32 @@ pub fn serialize<T: ?Sized>(value: &T, size_limit: SizeLimit) -> serde::Result<V
 /// encoding function, the encoder will verify that the structure can be encoded
 /// within that limit.  This verification occurs before any bytes are written to
 /// the Writer, so recovering from an error is easy.
-#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
-pub enum SizeLimit {
-    Infinite,
-    Bounded(u64)
+pub trait SizeLimit {
+    #[inline(always)]
```
dtolnay: My understanding is this doesn't do anything, the compiler just ignores it. The inline attribute needs to go where the body of the method is defined.

TyOverby: Oh dang.

TyOverby:
the serde-rs/bench test is a *really* micro-benchmark.

I think that I'm going to spend some time and make a benchmarking system that has a larger variety of tests.

PR #125, opened 2017-02-26
Inline all the things

TyOverby:
Depends on #124


TyOverby:
Turns out this slows things down.

PR #126, opened 2017-02-26
always preallocate

TyOverby:
depends on #125


TyOverby:
Closing in favor of #127

dgriffen:
Am I crazy? or is the diff of this the same as #127?

TyOverby:
@dgriffen This is a PR that depends on #125 (which depends on #124), while #127 only depends on #124 

For both of the PRs, look at the "branch target"

```
TyOverby  wants to merge 1 commit into inline-all-the-things from eager-prealloc
```
vs
```
TyOverby  wants to merge 1 commit into size-limit-trait from eager-prealloc-no-inline
```

PR #127, opened 2017-02-26
Eager prealloc no inline

TyOverby:
Always find the preallocated size


TyOverby:
Pushing in spite of travis failure because cargo nightlies are broken

PR #131, opened 2017-03-07
Make example work with serde

slyrz:
This PR makes the old example work with serde. It also replaces the asserts 
with some output of the encoded/decoded structs, because asserts are boring and an example needs some action.

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -1,35 +1,24 @@
-/*
+#[macro_use]
+extern crate serde_derive;
 extern crate bincode;
-extern crate
 
-use bincode::SizeLimit;
-use bincode::rustc_serialize::{encode, decode};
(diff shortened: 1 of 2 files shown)


dgriffen:
I think the asserts were useful because it demonstrated the compactness of the bincode format without having to run the example. With the new design, the user would need to run the example to determine the size of the encoded data.

TyOverby:
@slyrz: Ah, thanks for the PR, can't believe I missed these!  

Like @dgriffen said, I would prefer that the asserts are left as-is.  The examples are mainly supposed to be read, not executed.  (that said, they should still build and run).

TyOverby:
I do like the idea of printing the `World` at the end though.  (I'd just keep the asserts as well)

TyOverby:
Thanks!

PR #134, opened 2017-03-08
Make read_vec more robust

slyrz:
This PR makes bincode survive garbage input when decoding types that invoke
the `read_vec` function. Currently `read_vec` allocates a vector of the given
length without any sanity checks, which makes it easy to crash the decoder
with extreme values.

Since the `Read` trait does not provide a way to check the size of the 
remaining data, this PR splits `read_vec` into two functions:

1. `read_vec_small` that read vectors the way bincode currently does,
2. `read_vec_large` that tries to read largish vectors in chunks.

On encountering a large vector size, `read_vec` will call `read_vec_large`
to load the vector in reasonable sized chunks. `read_vec_large` avoids allocating
the vector at once and grows it instead. If length turns out to be garbage, 
bincode will reach the end of the data before it runs out of memory.

Of course, growing a buffer like this will come with a few extra allocations and 
copying, but with a reasonable large buffer size (I picked 64KB out of thin air), 
`read_vec_large` will rarely be invoked.

This PR is a mere suggestion. I don't know your stance on broken inputs and
how bincode should handle them. Some might say "Just add a checksum to your data"
and that's fine, so no feelings hurt if you don't want to merge this.

### DIFF
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -41,11 +41,7 @@ impl<R: Read, E: ByteOrder> Deserializer<R, E> {
     }
 
     fn read_bytes(&mut self, count: u64) -> Result<()> {
-        if let (read, false) = self.read.overflowing_add(count) {
-            self.read = read;
-        } else {
-            return Err(ErrorKind::SizeLimit.into());
-        }
+        self.read += count;
         match self.size_limit {
             SizeLimit::Infinite => Ok(()),
             SizeLimit::Bounded(x) if self.read <= x => Ok(()),
--- a/src/serde/reader.rs
+++ b/src/serde/reader.rs
@@ -41,7 +41,11 @@ impl<R: Read, E: ByteOrder> Deserializer<R, E> {
     }
 
     fn read_bytes(&mut self, count: u64) -> Result<()> {
-        self.read += count;
+        if let (read, false) = self.read.overflowing_add(count) {
+            self.read = read;
+        } else {
+            return Err(ErrorKind::SizeLimit.into());
+        }
         match self.size_limit {
             SizeLimit::Infinite => Ok(()),
             SizeLimit::Bounded(x) if self.read <= x => Ok(()),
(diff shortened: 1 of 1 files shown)


src/serde/reader.rs
```diff
@@ -54,15 +58,39 @@ impl<R: Read, E: ByteOrder> Deserializer<R, E> {
...
     }
 
+    fn read_vec_small(&mut self, len: usize) -> Result<Vec<u8>> {
+        let mut result = Vec::with_capacity(len);
+        unsafe { result.set_len(len); }
+        try!(self.reader.read_exact(&mut result));
+        Ok(result)
+    }
+
+    fn read_vec_large(&mut self, mut len: usize, block: usize) -> Result<Vec<u8>> {
```
TyOverby: This looks really good!

Shouldn't this optimization be possible while only maintaining one (result) buffer?

Here's my quick attempt: https://is.gd/Gq8N3C

src/serde/reader.rs
```diff
@@ -41,7 +41,11 @@ impl<R: Read, E: ByteOrder> Deserializer<R, E> {
     }
 
     fn read_bytes(&mut self, count: u64) -> Result<()> {
-        self.read += count;
+        if let (read, false) = self.read.overflowing_add(count) {
```
slyrz: My worry was a maliciously crafted length value causing an overflow panic. But you're right. People will have overflow checks disabled in their deployed release builds, which means an overflow won't cause their programs to terminate with a panic. I'll revert this. But be warned, once you add fuzzer tests to bincode, you'll find yourself adding this again. ;)

src/serde/reader.rs
slyrz: Yes, will change it. Your version needs another cmp::min in
```rust
let mut result = Vec::with_capacity(BLOCK_SIZE);
```
or else even the smallest Vectors reserve space for at least `BLOCK_SIZE` bytes. I'll try to get around this.

TyOverby: Oh wait, I think it's worse than that actually.  I think that it should read `Vec::new()`, because the very next thing that gets executed is a call to `reserve`.  

In my implementation I think I'm always over-allocating.

src/serde/reader.rs
TyOverby: Ah, good point, I see where this is happening.  Fortunately, I think there's a pretty easy solution:

Right now we are using read_bytes in two locations: 
* read_type
* read_vec

(I should probably rename one of those two methods, because despite the name, they are not similar at all).

`read_type` is safe, because it is only being used to increment the `read` counter for statically known types.
`read_vec` is not currently safe because it trusts the length that was read out of the byte-stream.  

This second case would be pretty easy to fix simply by sticking the calls to `read_bytes` inside of the `read_vec_small` and `read_vec_large` functions immediately after any calls to `read_exact`.

TyOverby:
Looks good!  Thanks for the changes!

PR #138, opened 2017-03-09
call read_bytes iteratively inside of read_vec

TyOverby:
@slyrz This should fix the overflow panic that you were talking about earlier


slyrz:
Looks good to me

PR #150, opened 2017-04-20
WIP Implement nocopy support for bincode

TyOverby:
TODO:

* [x] Find origina of test failure and fix it
* [x] Make deserialize_str and deserialize_bytes behave correctly in the face of a SizeLimit
* [x] Make BincodeRead public but prevent implementation
* [x] Stop using pub(crate)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ description = "A binary serialization / deserialization strategy that uses Serde
 [dependencies]
 byteorder = "1.0.0"
 num-traits = "0.1.32"
-serde = { git = "https://github.com/serde-rs/serde", branch = "master" }
+serde = "1.0.*"
 
 [dev-dependencies]
-serde_bytes = { git = "https://github.com/serde-rs/bytes", branch = "master" }
-serde_derive = { git = "https://github.com/serde-rs/serde", branch = "master" }
+serde_bytes = "0.10.*"
+serde_derive = "1.0.*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ description = "A binary serialization / deserialization strategy that uses Serde
 [dependencies]
 byteorder = "1.0.0"
 num-traits = "0.1.32"
-serde = "1.0.*"
+serde = "1.*.*"
 
 [dev-dependencies]
 serde_bytes = "0.10.*"
-serde_derive = "1.0.*"
+serde_derive = "1.*.*"
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -10,7 +10,7 @@ use ::SizeLimit;
 use super::{Result, Error, ErrorKind};
 use self::read::BincodeRead;
 
-pub(crate) mod read;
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -14,8 +14,8 @@ description = "A binary serialization / deserialization strategy that uses Serde
 [dependencies]
 byteorder = "1.0.0"
 num-traits = "0.1.32"
-serde = { git = "https://github.com/serde-rs/serde", branch = "master" }
+serde = "1.0.*"
```
dtolnay: This disallows upgrading to 1.1.0 which is not a breaking change. I think this should say "1.0".

src/de/mod.rs
```diff
@@ -8,6 +8,9 @@ use serde_crate::de::IntoDeserializer;
 use serde_crate::de::Error as DeError;
 use ::SizeLimit;
 use super::{Result, Error, ErrorKind};
+use self::read::BincodeRead;
+
+pub(crate) mod read;
```
dtolnay: `pub(crate)` doesn't work on stable yet so I would strongly discourage you from releasing with it.

dtolnay:
Other than that, and the checklist, I think that does it. Not so much code after all!

src/de/mod.rs
TyOverby: Ah yep, this was mainly added just to just so I wouldn't have to think about it when I was prototyping the new module structure.  I'll add this to my list of TODOs at the PR description.

TyOverby:
@dtolnay Yeah, I was pretty surprised at how easy it was!  (I think you helped out a lot with your initial 1.0 change though)

Cargo.toml
```diff
@@ -14,8 +14,8 @@ description = "A binary serialization / deserialization strategy that uses Serde
 [dependencies]
 byteorder = "1.0.0"
 num-traits = "0.1.32"
-serde = { git = "https://github.com/serde-rs/serde", branch = "master" }
+serde = "1.*.*"
```
cedenday: I don't like this, you should just write "^1", or more simply "1".

TyOverby: This is more explicit and easier to read.

TyOverby:
@dtolnay: I think this one is ready to merge in!

src/lib.rs
```diff
@@ -40,30 +40,32 @@ extern crate byteorder;
 extern crate num_traits;
 extern crate serde as serde_crate;
 
-mod serde;
+mod ser;
+mod de;
+pub mod internal;
```
dtolnay: This is not supposed to be public right?

TyOverby: I renamed `endian_choice` to `internal` and kept it public.  I'm debating if I should switch back to the old name though.

heartsucker:
Can you guys push this to crates.io? I'd like to be able to use this in my libs and not depend on git commits.

TyOverby:
@heartsucker pretty sure this is already published as 0.8.0

PR #191, opened 2017-07-07
Add Debug derivation to structs of the example in README.md

nelsonjchen:
The `assert_eq!` macro expects its arguments to satisfy the `Debug` trait as if and when it panics, it'll print out the `Debug` representation.


nelsonjchen:
I  guess this further completes the work in #187 which edited the example fully, but only a part of the README's copy.

TyOverby:
Oh thank you!

PR #195, opened 2017-07-19
Correctlly report the cause of IoError

### LINKED ISSUE #173: IO error cause
The `cause()` of `ErrorKind::IoError(err)` currently returns `err.cause()`. No, the cause is `err`!

dgriffen:
Closes #173


TyOverby:
Thanks!

PR #197, opened 2017-07-22
Add .vscode config directory to .gitignore

dgriffen:
This is useful for people who have workspace specific settings when editing with vscode. The most common of which is to turn off format on save since the current code base isn't formatted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.swp
 *.swo
 .cargo
+.vscode


dtolnay:
Check out [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). That seems easier than adding this to gitignore of individual projects.

TyOverby:
agree with @dtolnay

PR #198, opened 2017-07-24
Reject trailing bytes

### LINKED ISSUE #193: Reject trailing bytes when deserializing from &[u8]
In `deserialize_from`, it makes sense to allow deserializing multiple values from the same io::Read. But in `deserialize`, I think trailing data should be an error.

I would expect this to be an error rather than ignoring `b"x"`:

```rust
extern crate bincode;

fn main() {
    println!("{:?}", bincode::deserialize::<char>(b"1x"));
}
```

ZoeyR:
Closes #193 

There were two design possibilities I considered for tackling this problem. The current one changes the impls on `SliceReader` to be impls for `&mut SliceReader`. This allows me to access it after passing it to the deserializer. The other option would be to allow `pub(restricted)` access to the reader field of the deserializer and access it from there.

### DIFF
--- a/src/de/read.rs
+++ b/src/de/read.rs
@@ -68,7 +68,7 @@ impl<R> IoReader<R> {
     }
 }
 
-impl<'a, 'storage> io::Read for &'a mut SliceReader<'storage> {
+impl<'storage> io::Read for SliceReader<'storage> {
     #[inline(always)]
     fn read(&mut self, out: &mut [u8]) -> io::Result<usize> {
         if out.len() > self.slice.len() {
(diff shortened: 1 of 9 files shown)


ZoeyR:
Finally got around to updating this to current bincode after 2.5 years 😅

src/de/read.rs
```diff
@@ -101,7 +105,7 @@ impl<'storage> SliceReader<'storage> {
     }
 }
 
-impl<'storage> BincodeRead<'storage> for SliceReader<'storage> {
+impl<'a, 'storage> BincodeRead<'storage> for &'a mut SliceReader<'storage> {
```
jdm: Why are these changes necessary?

ZoeyR: I guess now that we use `pub(crate)` throughout the coebase I can use the other option I outlined in my PR. Guess that happens when you don't update a PR for two years lol

ZoeyR:
I've realized that this might be considered a breaking change.

strega-nil:
I was the person who was saying that this should be considered a breaking change, fwiw.

ZoeyR:
cc @strega-nil

PR #199, opened 2017-07-24
Internal type names

### LINKED ISSUE #163: Do not use IoRead or IoError in public API
The rustdoc for BincodeRead currently looks like:

```rust
pub trait BincodeRead<'storage>: IoRead { }
```

IoRead is a renamed import that you use internally. Since this is user-facing, it should use the standard names for std types.

dgriffen:
Closes #163 

I also took the opportunity to rename `IoReadReader` to the less redundant `IoReader`.


TyOverby:
Looks good!  Thanks!

PR #200, opened 2017-07-24
take reader and writer by value

### LINKED ISSUE #159: Reader and writer should be taken by value
```rust
pub fn deserialize_from<R: ?Sized, T, S>(reader: &mut R, size_limit: S)

pub fn serialize_into<W: ?Sized, T: ?Sized, S>(writer: &mut W, value: &T, size_limit: S)
```

This is obnoxious when reading or writing some IO stream that is used only for bincode.

```rust
bincode::deserialize_from(File::open(f)?, infinite)
```

The standard library implements Read for &mut R and Write for &mut W so if these methods take the argument by value, callers can keep passing &mut like before if they want.

dgriffen:
Closes #159


TyOverby:
Great, thanks!

PR #203, opened 2017-08-18
[WIP] format code with rustfmt

dgriffen:
I ran the code here through `rustfmt`. There may be errors in the way the code was automatically formatted that makes it inconsistent with the style guides.


TyOverby:
Since running rustfmt is so easy, I'm going to close this and do it myself rather than resolve conflicts.

Issue #206, opened 2017-10-09
deserialization code leaves much to be desired


jrmuizel:
The following rust code gives us a lot of per field overhead:
```rust
#[derive(Deserialize)]
pub struct P {
       x: u32,
       y: u32,
       z: u32,
       w: u32
}

pub fn moo(encoded: Vec<u8>) -> P {
    let mut s :&[u8] = &encoded;
    deserialize_from(&mut s, bincode::Infinite).unwrap()
}
```

I'd ideally want this to compile to some length checks and a single memcpy for all of the fields. Instead I get something that looks like a memcpy loop per field.

I suspect there are a number of things contributing to the problem here. The full assembly of moo follows:

```assembly
	.globl	__ZN11serde_deser3moo17hd4391294d14f34bdE
	.p2align	4, 0x90
__ZN11serde_deser3moo17hd4391294d14f34bdE:
	.cfi_startproc
	pushq	%rbp
Lcfi16:
	.cfi_def_cfa_offset 16
Lcfi17:
	.cfi_offset %rbp, -16
	movq	%rsp, %rbp
Lcfi18:
	.cfi_def_cfa_register %rbp
	pushq	%r15
	pushq	%r14
	pushq	%r13
	pushq	%r12
	pushq	%rbx
	subq	$72, %rsp
Lcfi19:
	.cfi_offset %rbx, -56
Lcfi20:
	.cfi_offset %r12, -48
Lcfi21:
	.cfi_offset %r13, -40
Lcfi22:
	.cfi_offset %r14, -32
Lcfi23:
	.cfi_offset %r15, -24
	movq	%rdi, -112(%rbp)
	movq	(%rsi), %r15
	movq	8(%rsi), %rax
	movq	%rax, -104(%rbp)
	movq	16(%rsi), %r12
	leaq	-68(%rbp), %r13
	movl	$0, -68(%rbp)
	movl	$4, %ebx
	movq	%r15, -80(%rbp)
	.p2align	4, 0x90
LBB4_1:
	cmpq	%rbx, %r12
	movq	%r12, %r14
	cmovaeq	%rbx, %r14
	cmpq	$1, %r14
	jne	LBB4_3
	movzbl	(%r15), %eax
	movb	%al, (%r13)
	jmp	LBB4_4
	.p2align	4, 0x90
LBB4_3:
	movq	%r13, %rdi
	movq	%r15, %rsi
	movq	%r14, %rdx
	callq	_memcpy
	testq	%r14, %r14
[...]

PR #207, opened 2017-10-10
Forward read_exact() as well as read().

jrmuizel:
If we don't do this we end up using the generic read_exact method
which is not necessarily optimal. This is especially when
using a specialized Read implementation to go fast.
See https://github.com/TyOverby/bincode/issues/206


TyOverby:
Looks good, thanks!

I'm also really interested in your specialized Read implementation.  Is this the one that eschews bounds checking?

pcwalton:
@TyOverby It's at https://github.com/servo/webrender/pull/1834.

jrmuizel:
I just realized that this pull request alone will improve deserialization performance because we'll now be using the read_exact() implementation on &mut [u8] which is simpler than the default read() implementation.

PR #214, opened 2017-10-17
2nd implementation of the Config Trait

### LINKED ISSUE #156: Rename or hide bincode::internal
It is confusing to have this module with a ton of public functionality in it. Am I allowed to use it? Is it reserved for internal bincode use? What makes it internal? Why is it public and visible in the docs?

### LINKED ISSUE #160: Alternative design for SizeLimit
It looks like the large majority of Bincode users on GitHub are using Infinite as the size limit. It would be nice to redesign the API to accommodate this use case more ergonomically.

### LINKED ISSUE #164: Serialized size is not affected by endian choice
The `serialized_size` and `serialized_size_bounded` functions should not be exported under `endian_choice`. They have nothing to do with it.

TyOverby:
closes #160 
closes #156 
closes #164

Basic gist of it is that you can write the following 

```rust
use bincode::{serialize, config};

serialize(&vec![1,2,3]);
// or 
config().big_endian().limit(12).serialize(&vec![1,2,3]);
```

depending on if you want to configure bincode.

Adding new configuration options should be backwards compatible.  All of this should be 0 cost (assuming that the inliner does its work correctly)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,7 @@
 use super::internal::{Bounded, Infinite, SizeLimit};
 use ::error::Result;
 use byteorder::{BigEndian, ByteOrder, LittleEndian, NativeEndian};
+use {DeserializerAcceptor, SerializerAcceptor};
 use serde;
 use std::io::{Write, Read};
 use std::marker::PhantomData;
@@ -265,10 +266,32 @@ impl Config {
     /// Deserializes an object from a custom `BincodeRead`er using the default configuration.
     /// It is highly recommended to use `deserialize_from` unless you need to implement
     /// `BincodeRead` for performance reasons.
-    /// 
+    ///
     /// If this returns an `Error`, `reader` may be in an invalid state.
     #[inline(always)]
     pub fn deserialize_from_custom<'a, R: BincodeRead<'a>, T: serde::de::DeserializeOwned>(&self, reader: R) -> Result<T> {
         config_map!(self, opts => ::internal::deserialize_from_custom(reader, opts))
+    }
+
+    /// Executes the acceptor with a serde::Deserializer instance.
+    pub fn with_deserializer<'a, A,  R>(&self, reader: R, acceptor: A) -> A::Output
+    where A: DeserializerAcceptor<'a>,
+          R: BincodeRead<'a>
+    {
+        config_map!(self, opts => {
+            let mut deserializer = ::de::Deserializer::new(reader, opts);
+            acceptor.accept(&mut deserializer)
+        })
+    }
+
+    /// Executes the acceptor with a serde::Serializer instance.
+    pub fn with_serializer<A, W>(&self, writer: W, acceptor: A) -> A::Output
+    where A: SerializerAcceptor,
+        W: Write
+    {
+        config_map!(self, opts => {
+            let mut serializer = ::ser::Serializer::new(writer, opts);
+            acceptor.accept(&mut serializer)
+        })
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,64 @@
+use super::{SizeLimit, Infinite, Bounded};
+use byteorder::{ByteOrder, BigEndian, LittleEndian, NativeEndian};
+use std::marker::PhantomData;
+
+trait Options {
+    type Limit: SizeLimit + 'static;
+    type Endian: ByteOrder + 'static;
+
+    fn limit(&mut self) -> &mut Self::Limit;
+}
+
+pub trait Config {
+    fn with_no_limit(self: Box<Self>) -> Box<Config>;
+    fn with_limit(self: Box<Self>, u64) -> Box<Config>;
+    fn with_little_endian(self: Box<Self>) -> Box<Config>;
+    fn with_big_endian(self: Box<Self>) -> Box<Config>;
+    fn with_native_endian(self: Box<Self>) -> Box<Config>;
+}
+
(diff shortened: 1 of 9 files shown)


TyOverby:
@dtolnay: this is a rewrite of my older PR with some suggestions from you taken into account

src/config.rs
```diff
@@ -0,0 +1,266 @@
...
+use self::LimitOption::*;
+use self::EndianOption::*;
+
+struct DefaultOptions(Infinite);
+
+pub(crate) trait Options {
+    type Limit: SizeLimit + 'static;
+    type Endian: ByteOrder + 'static;
+
+    #[inline(always)]
```
dtolnay: I think the inline attribute only means anything on functions that have a body. I don't think this automatically applies to all implementations of the trait (but I could be wrong).

TyOverby: Done.

src/config.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub(crate) fn new() -> Config {
+        Config {
+            limit: LimitOption::Unlimited,
+            endian: EndianOption::Little,
+        }
+    }
+
+    /// TODO: Document
+    #[inline(always)]
+    pub fn with_no_limit(self) -> Config {
```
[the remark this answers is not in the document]
TyOverby: Done.

src/config.rs
```diff
@@ -0,0 +1,266 @@
...
+        Config {
+            limit: LimitOption::Unlimited,
+            ..self
+        }
+    }
+
+    /// TODO: Document
+    #[inline(always)]
+    pub fn with_limit(self, limit: u64) -> Config {
```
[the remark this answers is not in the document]
TyOverby: Done.

TyOverby:
@dtolnay: any more feedback on the design?  

If not, I'm going to fill out the doc comments and merge.

TyOverby:
Oh, one issue is that I'm no longer exporting BincodeRead, so people that relied on that (servo) will break.

I wonder if I could incorperate their custom BincodeRead into bincode and provide an unsafe fn for using it.

src/config.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    /// TODO: Document
+    #[inline(always)]
+    pub fn serialized_size<T: ?Sized + serde::Serialize>(&self, t: &T) -> u64 {
+        config_map!(self.limit, self.endian, opts => ::internal::serialized_size(t, opts))
+    }
+
+    /// TODO: Document
+    #[inline(always)]
+    pub fn serialized_size_bounded<T: ?Sized + serde::Serialize>(&self, t: &T, bound: u64) -> Option<u64> {
```
dtolnay: I haven't thought through which way is better, but did you consider dropping this method and changing `serialized_size` to return a Result and then recommending `bincode::config().limit(bound).serialized_size(&t)`?

src/config.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    /// TODO: Document
+    #[inline(always)]
+    pub fn serialize<T: ?Sized + serde::Serialize>(&self, t: &T) -> Result<Vec<u8>> {
+        config_map!(self.limit, self.endian, opts => ::internal::serialize(t, opts))
+    }
+
+    /// TODO: Document
+    #[inline(always)]
+    pub fn serialized_size<T: ?Sized + serde::Serialize>(&self, t: &T) -> u64 {
```
dtolnay: I think this signature with u64 instead of Result is fine but the documentation definitely needs to explain what happens if serialization fails.

On the other hand there is some value in exposing that serialization failed in `serialized_size` before the caller goes on to invoke `serialize_into` which may allocate a bunch of memory and do a bunch of work. I left a comment on this below.

src/internal.rs
```diff
@@ -138,51 +17,45 @@ impl serde::ser::Error for Error {
...
         Some(size_limit) => {
-            let actual_size = try!(serialized_size_bounded(value, size_limit).ok_or(
+            let actual_size = try!(serialized_size_bounded(value, size_limit, &mut options).ok_or(
                 ErrorKind::SizeLimit,
             ));
             Vec::with_capacity(actual_size as usize)
         }
         None => {
-            let size = serialized_size(value) as usize;
+            let size = serialized_size(value, &mut options) as usize;
```
dtolnay: Whether or not we change to Result in the public `serialized_size`, I think this should return early if `serialized_size` failed.

src/config.rs
TyOverby: No effect.

dtolnay: Let's remove it then. It is still present in the most recent revision.

src/config.rs
TyOverby: That makes way too much sense.

src/config.rs
TyOverby: Nah, I think that Result<u64> is a pretty good move.

src/internal.rs
TyOverby: Agreed.  This has been changed to short circuit

src/config.rs
TyOverby: Oh yeah, I accidentally removed them from the default methods on OptionsExt instead.  I wonder if they are used on default methods?

I'll look into this a bit more.

TyOverby:
Pretty much done with this PR.  I added the `deserialize_from_custom` method, and updated the docs pretty substantially.

I'm going to talk with servo to see if this meets their needs and then I'll hit merge.

TyOverby:
I talked to @Gankro from servo and I'm pretty sure that with the `with_deserializer` method, we'll be able to hit all their requirements.

PR #220, opened 2017-11-13
do not exclude LICENSE.md

ignatenkobrain:
MIT license requires that it is shipped with sources. So crates.io archive should have it as well.


TyOverby:
Thanks!

PR #222, opened 2017-11-29
bump byteorder version to get transmuting binary float parses

Gankro:
this is faster, and produces more accurate (bit-exact) round-tripping.


TyOverby:
Neat!

PR #223, opened 2017-12-01
Elide unnecessary copy when consuming a byte buffer.

khuey:
The byte buffer stuff is designed to allow zero copies ... but bincode forces one anyways, to save reallocating its internal `Vec`.


TyOverby:
Good catch!  Thanks!

PR #228, opened 2018-02-26
Update README example to match #214 changes

iKevinY:
As shown in `examples/basic.rs`, `bincode::Infinite` isn't used anymore.


TyOverby:
Good catch, thanks!

PR #237, opened 2018-05-30
Add support for 128bit numbers

### LINKED ISSUE #236: Support 128bit integers
`serde` ~~`1.0.62`~~ `1.0.60` was recently released with optional support for 128bit numbers. It would be good if we could wire them up to `bincode`. There's a convenient `serde_if_integer128!` macro that `serde` exposes we could use to automatically detect newer compilers, but that might not line up so well with `byteorder`'s explicit `i128` feature. Maybe we could do something like this:

In `Cargo.toml`:

```toml
[features]
i128 = ["byteorder/i128"]
```

In `ser.rs` or `de.rs`:

```rust
impl Serializer {
    serde_if_integer128! {
        #[cfg(feature = "i128")]
        fn serialize_i128(self, v: i128) -> Result<Self::Ok, Self::Error> {
            // use `byteorder` to serialize our value
        }

        #[cfg(not(feature = "i128"))]
        fn serialize_i128(self, v: i128) -> Result<Self::Ok, Self::Error> {
            let _ = v;
            Err(Error::custom("i128 is not supported. Enable the `i128` feature of `bincode`"))
        }

        ...
    }
}
```

so that vanilla `bincode` keeps working on older compilers, but has a more helpful error message if you accidentally use 128bit numbers without enabling the feature?

Or maybe we could work around it using some `build.rs` wizardry?

KodrAus:
Closes #236 

This PR adds support for encoding/decoding `i128` and `u128` which were stabilized in `rustc` `1.26.0`, and supported as of `serde` `1.0.60`.

There's a new a disabled-by-default `i128` feature flag in `bincode`, which enables support for 128bit numbers. If `bincode` is compiled on a version of Rust that supports 128bit numbers but doesn't have the `i128` feature enabled then we'll return a custom error at runtime suggesting they enable the feature.

If there's anything you'd like changed let me know. Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,12 @@ rust:
     - beta
     - nightly
 
+matrix:
+  include:
+  - rust: 1.18.0
+    script:
+    - cargo test
+
 script:
     - cargo test
     - cargo test --features "i128"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,7 @@ rust:
     - stable
     - beta
     - nightly
+
+script:
+    - cargo test
+    - cargo test --features "i128"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,11 @@ description = "A binary serialization / deserialization strategy that uses Serde
 
 [dependencies]
 byteorder = "1.2.0"
-serde = "^1.0.27"
+serde = "^1.0.63"
 
 [dev-dependencies]
 serde_bytes = "^0.10.3"
 serde_derive = "^1.0.27"
 
+[features]
+i128 = ["byteorder/i128"]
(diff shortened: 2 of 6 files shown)


KodrAus:
cc @TyOverby @dtolnay

KodrAus:
I could rejig the `.travis.yml` a bit so we just build without the feature, and test with it so there's less output to worry about? Would you be interested in pinning a specific version of Rust in CI so we don't accidentally bump it when making changes? I was testing these changes locally on the latest stable and `1.18.0`?

TyOverby:
This looks great!  Thanks!

My only request would be adding a test that utilizes the full range of the 128 bit numbers.

TyOverby:
I'm totally fine with the travis output spew as I rarely need to look at it :D 

As for pinning the version, I don't know?  It looks like serde tests with 

  - 1.13.0
  - 1.15.0
  - stable
  - beta
- nightly

If you'd like, feel free to build these into this PR (in a new commit) and we can enable testing on these channels.

KodrAus:
@TyOverby It looks like we need `1.18.0` for the `pub(restricted)` syntax, so I've added that to the CI.

TyOverby:
Looks great, thanks for the contributions!

PR #246, opened 2018-09-03
Introduce `Config::deserialize_seed` for exposing serde's statefull deserialization.

boxdot:
This is needed if one wants to use stateful deserialization, since the `SliceReader` is public but not part of the public API, and `bincode::Deserializer` is private.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -266,6 +266,12 @@ impl Config {
         config_map!(self, opts => ::internal::deserialize_in_place(reader, opts, place))
     }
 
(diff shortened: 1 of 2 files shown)


TyOverby:
Looks good!  Could you add a few tests?  Maybe one test with the default configuration and one with a non-default configuration?

boxdot:
Makes sense, tests added. -- Btw., it looks like the source code of bincode is not formatted at all. Probably, this could be done to make it easier for people who are using `rustfmt`.

TyOverby:
I just formatted the code!

PR #261, opened 2019-01-28
Make i128 support automatic for supporting Rustc versions

### LINKED ISSUE #250: Make `i128` support automatic
To revisit #236, it would be more convenient for users if `i128` support was enabled automatically on compatible compilers. We [already argued in July](https://github.com/rust-random/rand/pull/571) that this was the best option for Rand.

The linked PR shows how this can be achieved without any breakage (except for old nightlies, which are generally not possible to ensure support for anyway).

dhardy:
Fix #250.

Uses autocfg. If you prefer, I can avoid this as @BurntSushi [did in byteorder](https://github.com/dhardy/byteorder/blob/master/build.rs), but this seems like an apt tool for the job so worth making more popular.

Tested on Rustc 1.22.1, 1.31.1 and a recent nightly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ documentation = "https://docs.rs/bincode"
 readme = "./readme.md"
 categories = ["network-programming"]
 keywords = ["binary", "encode", "decode", "serialize", "deserialize"]
+build = "build.rs"
 
 license = "MIT"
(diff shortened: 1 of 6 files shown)


TyOverby:
When I get home I'll release this!  

To confirm: this is a minor-release bump correct?

dhardy:
Yes, it shouldn't have any breaking changes (hence keeping the dummy feature).

PR #268, opened 2019-05-15
Fix autocfg version

quininer:
`autocfg` 0.1.1 does not have `rerun_path`, which causes it to compile errors.


dtolnay:
Published in 1.1.4.

PR #271, opened 2019-06-08
Varint enum tags and lengths behind a feature

### LINKED ISSUE #75: Consider encoding str/seq lengths as varints
Have you considered storing the lengths of strings and sequences as [Varints](https://developers.google.com/protocol-buffers/docs/encoding#varints)? The fixed 8-byte encoding of the length adds quite some space overhead when you store many shorter strings.

Just an idea. I could create a pull request if you like it.

### LINKED ISSUE #157: Variable length encoding for enum discriminants
I think the majority of enum discriminants are less than 128. A very simple variable length encoding like encode \<128 as one byte and >=128 as four bytes would address most concerns from people who are astonished that Bincode encodes their 2-variant enum with `[0x00, 0x00, 0x00, 0x00]` and `[0x00, 0x00, 0x00, 0x01]` tags.

maciejhirsz:
Resolves #157, resolves #75.

Varint support needs to be enabled with the `varint` feature, it only affects enum tags and sequence lengths, all fixed size ints travel over the wire without alterations.

The particular encoding is pretty straight forward: while `n` is greater then `127` keep taking lowest 7 bits and serialize them as `u8` with significant bit set to `1`, final byte is always encoded without significant bit. Varint ignores the endianess setting, partly because I think it's unnecessary in this case, partly because I didn't want to wrap my head around it and potentially introduce weird cases where the first byte for a value `>= 128` lacks significant bit set. I didn't bother to check which direction ProtoBuf or Go varint does things, if it's different I can change it.

#167 can still technically be done as I don't verify that varints are encoded to their smallest possible representation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ rust:
     - beta
     - nightly
 
-script: cargo test
+script:
+  - cargo test
+  - cargo test --features "varint"
 
 matrix:
   include:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,5 +33,9 @@ autocfg = "0.1.2"
 # feature will be removed if and when a new major version is released.
 i128 = []
 
+# Enabling this feature will make bincode treat enum discriminants and
+# sequence lengths as varints instead of u32s.
+varint = []
+
 [badges]
 travis-ci = { repository = "servo/bincode" }
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -48,7 +48,7 @@ impl<'de, R: BincodeRead<'de>, O: Options> Deserializer<R, O> {
     }
 
     fn read_vec(&mut self) -> Result<Vec<u8>> {
-        let len: usize = try!(serde::Deserialize::deserialize(&mut *self));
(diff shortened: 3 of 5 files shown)


maciejhirsz:
I think it should be possible to use the `Options` trait for this, instead of a feature flag, with no extra runtime cost. The advantage would be that this setting isn't suddenly per-project global (which might matter if you use bincode for IPC and for transmiting data over the wire, with both of those having different requirements). On the flip side, that would be a breaking change (albeit only on the API surface, not affecting the ability to read old data).

maciejhirsz:
I was afraid we'd have to add an enum variant or some such, but looking at how `Config` is constructed right now, it should be pretty straight forward to just add a method there. I'll have a go at it.

Tolsi:
Any updates?

maciejhirsz:
@Tolsi sorry, I haven't put more time into this since, if anyone wants to pick it up from where I left it feel free, I don't think I'll have time to work on this anytime soon.

ZoeyR:
Closing in favor of #306

PR #273, opened 2019-07-22
Add `Clone` + `PartialEq` traits to `bincode::ErrorKind`

SamuelMarks:
Add [`Clone`](https://doc.rust-lang.org/std/clone/trait.Clone.html) trait for [`bincode::ErrorKind`](https://github.com/servo/bincode/blob/72bb6ef/src/error.rs#L16-L40). Also add [`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) trait to allow use [`assert_eq!`](https://doc.rust-lang.org/std/macro.assert_eq.html) in tests.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -39,6 +39,66 @@ pub enum ErrorKind {
     Custom(String),
 }
 
+impl Clone for ErrorKind {
+    fn clone(&self) -> ErrorKind {
+        match self {
+            ErrorKind::Io(ioe) => {
+                ErrorKind::Io(std::io::Error::new(ioe.kind(), format!("{:?}", ioe)))
+            }
+            ErrorKind::InvalidUtf8Encoding(utferror) => {
+                ErrorKind::InvalidUtf8Encoding(utferror.clone())
+            }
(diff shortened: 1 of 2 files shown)


jdm:
It's not clear to me why the 1.18 build for Travis CI is failing. I'm also not thrilled by the idea of a PartialEq implementation that lies about Io errors

SamuelMarks:
It's difficult to decide between ways of handling errors, and creating large Rust projects that can propagate errors of any type.

Happy to make changes to get it approved by you. Tell us what to do.

FYI: That IO line is same as https://github.com/spacejam/sled/blob/3b20bf4/crates/pagecache/src/result.rs#L103

jdm:
Unless there is a stronger motivation for PartialEq than writing tests, I would prefer not to include that change.

dtolnay:
I would prefer to pass on the Clone impl as well. In code that needs multiple owned values of the same error, it should use Rc\<bincode::Error\> or similar. As an advantage over the impl in this PR, that approach preserves the cause of the io error. It would be surprising for Clone to produce an error that behaves differently from the original.

PR #282, opened 2019-10-26
Added Clone impl to Config

jean-airoldie:


nox:
I would remove `Copy`, as stuff can be added to the config structure later and if it can't be `Copy` anymoer we will need a breaking change.

jean-airoldie:
Sure thing.

PR #284, opened 2019-10-30
improve safety of fill_buffer

### LINKED ISSUE #260: UB in IoReader::fill_buffer
While doing an unsafe-audit of bincode, I found this code in `src/de/read.rs`:

```rust
impl<R> IoReader<R>
where
    R: io::Read,
{
    fn fill_buffer(&mut self, length: usize) -> Result<()> {
        let current_length = self.temp_buffer.len();
        if length > current_length {
            self.temp_buffer.reserve_exact(length - current_length);
        }

        unsafe {
            self.temp_buffer.set_len(length);
        }

        self.reader.read_exact(&mut self.temp_buffer)?;
        Ok(())
    }
}
```

If read_exact errors, this code might leave temp_buffer longer than the number of initialized bytes. Also, passing uninitialized memory to read_exact is UB. Since `read_exact` is a safe fn, any slice that's passed in must be initialized.

From @rkruppe on Zulip:
> I don't think it is [safe]. If `self.reader.read_exact` doesn't fill the buffer (e.g. there is an I/O error while reading, or it's implemented incorrectly, or it panics for some reason) then this code exposes uninitialized bytes to the world. And while it's still up for debate whether there are *some* things that can be done in careful unsafe code with uninitialized bytes, it's certain that uninitialized memory can lead to UB when used in unsuspecting safe code.
>
> It also exposes uninitialized memory to the implementation of `read_exact` so if that reads from the buffer (which it "shouldn't" but safe code can do it) it'll get uninitialized bytes too, with the same consequences.

JoshMcguigan:
Fixes #260 

This implements a fix for a safety issue found in `fill_buffer` as reported in #260. This is the implementation suggested by @Dr-Emann.

### DIFF
--- a/src/de/read.rs
+++ b/src/de/read.rs
@@ -1,6 +1,6 @@
 use error::Result;
 use serde;
-use std::io;
+use std::{io, slice};
 
 /// An optional Read trait for advanced Bincode usage.
 ///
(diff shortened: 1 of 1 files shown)


jdm:
Thank you!

PR #294, opened 2019-12-13
Issue #285: Add support for lengths other than 64 bits.

### LINKED ISSUE #285: Support non-64bit lengths.
In the documentation is says:
```
str is encoded as (u64, &[u8]), where the u64 is the number of bytes contained in the encoded string.
```

I am looking for a way around this to explicitly control how the length is serialized. This is similar to: https://github.com/servo/bincode/issues/182
and 
https://github.com/servo/bincode/issues/238

I could not find a good way to implement this even manually. I can create a custom type and implement Serialize, but calling `serialize_bytes()` adds a length prefix and there doesn't appear to be any other way to induce a call to `write_all()`.

Would it be possible to do this by adding a Bincode specific trait extending Serialize perhaps? 
If that can be solved Deserialization seems like it should be straight forward.

tkaitchuck:
This change adds two new config values `string_size` and `array_size`. 
Both of which take an enum to allow the lengths of strings and vectors/arrays to be set to:
`U8`, `U16`, `U32`, and `U64`. Where `U64` is the default. (Which is compatible with current behavior.)

In the event that the data being serialized does not fit into the size specified the serialization will fail with `ErrorKind::SizeTypeLimit`

Fixes #285 .

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,4 @@
-use super::internal::{Bounded, Infinite, SizeLimit, SizeType};
+use super::internal::{Bounded, Infinite, SizeLimit, SizeType, U8, U16, U32, U64};
 use byteorder::{BigEndian, ByteOrder, LittleEndian, NativeEndian};
 use de::read::BincodeRead;
 use error::Result;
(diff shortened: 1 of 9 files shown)


tkaitchuck:
@dtolnay Does this look like a good approach?

tkaitchuck:
Moved here: https://github.com/pravega/bincode2/pull/4

PR #302, opened 2020-01-30
minor cleanup; more idiomatic example/ code

Byron:
While looking into bincode, I thought I could contribute some minor improvements. Thanks for making this fantastic library!

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -13,15 +13,16 @@ struct Entity {
 #[derive(Serialize, Deserialize, PartialEq, Debug)]
 struct World(Vec<Entity>);
 
-fn main() {
+fn main() -> Result<(), Box<dyn std::error::Error>>{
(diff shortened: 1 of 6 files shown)


jdm:
So apparently rust 1.18 doesn't like the `&dyn` syntax. That's a version of Rust that was released in 2017; I wonder if it's time to update the minimum supported Rust version yet...

Byron:
Oh, that is surprising! I would have expected the (recent) Rust compiler to not complain about things prior to edition 2018.

If this project wishes to maintain Rustc v1.18 compatibility, it's probably best to close this PR. 

Alternatively it would be worth re-evaluating which Rust version now is suitable - it's probably depending mostly on this crate's consumers.

Please feel free to do whatever you find appropriate.

ZoeyR:
For 1.0 crates I've always been of the opinion that bumping the msrv is a breaking change.

Byron:
I am closing the PR as it probably can't be merged in the current form.

PR #306, opened 2020-03-14
Varint enum tags and lengths

### LINKED ISSUE #75: Consider encoding str/seq lengths as varints

### LINKED ISSUE #157: Variable length encoding for enum discriminants

### LINKED ISSUE #319: Varint encoding for discriminants/lengths
Rolling up issues #285, #157, and #75

We should offer an option to allow encoding discriminants and lengths as some sort of varint.

ZoeyR:
Resolves #319 

This is a continuation of PR #271 based on feedback in that PR. The primary change is to use a config option instead of a feature in order to use varint sequence lengths and discriminants.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,5 @@ serde_derive = "1.0.27"
 # major version is released.
 i128 = []
 
-# Enabling this feature will make bincode treat enum discriminants and
-# sequence lengths as varints instead of u32s.
-varint = []
-
 [badges]
 travis-ci = { repository = "servo/bincode" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,5 +30,9 @@ serde_derive = "1.0.27"
 # major version is released.
 i128 = []
 
+# Enabling this feature will make bincode treat enum discriminants and
+# sequence lengths as varints instead of u32s.
+varint = []
+
 [badges]
 travis-ci = { repository = "servo/bincode" }
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,6 +4,7 @@ use error::{ErrorKind, Result};
 use serde;
 use std::io::{Read, Write};
 use std::marker::PhantomData;
+use std::mem::size_of;
 
 pub(crate) use self::internal::*;
 
@@ -272,21 +273,21 @@ pub struct FixedLength;
 #[derive(Copy, Clone)]
 pub struct VarintLength;
 
-impl VarintLength {
-    const SINGLE_BYTE_MAX: u8 = 250;
-    const U16_BYTE: u8 = 251;
-    const U32_BYTE: u8 = 252;
-    const U64_BYTE: u8 = 253;
+const SINGLE_BYTE_MAX: u8 = 250;
+const U16_BYTE: u8 = 251;
+const U32_BYTE: u8 = 252;
+const U64_BYTE: u8 = 253;
 
+impl VarintLength {
     fn varint_size(n: u64) -> u64 {
-        if n <= Self::SINGLE_BYTE_MAX as u64 {
+        if n <= SINGLE_BYTE_MAX as u64 {
             1
         } else if n <= u16::max_value() as u64 {
-            (1 + std::mem::size_of::<u16>()) as u64
+            (1 + size_of::<u16>()) as u64
         } else if n <= u32::max_value() as u64 {
-            (1 + std::mem::size_of::<u32>()) as u64
+            (1 + size_of::<u32>()) as u64
         } else {
-            (1 + std::mem::size_of::<u64>()) as u64
+            (1 + size_of::<u64>()) as u64
         }
     }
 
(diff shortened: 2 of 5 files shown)


ZoeyR:
@dtolnay can I get a review of this?

ZoeyR:
@dtolnay I thought based on #157 (which you opened) and #271 (which you seemed agreeable towards) that this PR would be a good fit for the library. It doesn't offer all of the configuration of #294 because IMO the amount of config in #294 is a bit of a footgun.

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+        }
+
+        bytes
+    }
+
+    fn serialize_varint<W: Write, O: Options>(ser: &mut ::ser::Serializer<W, O>, mut n: usize) -> Result<()> {
+        use serde::Serialize;
+
+        while n > 127 {
+            (128 | n as u8).serialize(&mut *ser)?;
```
strega-nil: Also, could you use binary constants here instead of `128`? It'd be even better if you added a named constant.

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+    fn deserialize_varint<'de, R: BincodeRead<'de>, O: Options>(de: &mut ::de::Deserializer<R, O>) -> Result<usize> {
+        use std::mem::size_of;
+
+        let mut n = 0;
+        let mut shift = 0;
+        let mut byte: u8 = serde::Deserialize::deserialize(&mut *de)?;
+
+        // Only allow reading size_of + 1 bytes to avoid overflows on bitshifts
+        for _ in 0..(size_of::<usize>()) {
+            if byte < 128 {
```
strega-nil: checking `byte < 128` is more confusing than `byte & 0b1000_0000 == 0`, imo.

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+                break;
+            }
+
+            n |= ((byte & 127) as usize) << shift;
```
strega-nil: same as above, use binary or named constants.

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+    }
+
+    fn deserialize_varint<'de, R: BincodeRead<'de>, O: Options>(de: &mut ::de::Deserializer<R, O>) -> Result<usize> {
+        use std::mem::size_of;
+
+        let mut n = 0;
+        let mut shift = 0;
+        let mut byte: u8 = serde::Deserialize::deserialize(&mut *de)?;
+
+        // Only allow reading size_of + 1 bytes to avoid overflows on bitshifts
```
strega-nil: This is not the case. You should be checking the actual value of the result (although it's totally fine to add the additional constraint that the varint is encoded in `ceil(8 * size_of / 7)` bytes). Additionally, you don't actually check that the last byte is a final byte!

This will break in the case of the following byte stream, on `usize = u32`:

```
deserialize_varint(< 0xFF, 0xFF, 0xFF, 0xFF, 0x7F >)
```

It will attempt to shift `0x7F` by `7 * 4  = 28`, which will result in an overflow.

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+            }
+
+            n |= ((byte & 127) as usize) << shift;
+            shift += 7;
+            byte = serde::Deserialize::deserialize(&mut *de)?;
+        }
+
+        Ok(n | ((byte as usize) << shift))
+    }
+}
```
strega-nil: Add a newline here

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+
+        while n > 127 {
+            n >>= 7;
+            bytes += 1;
+        }
+
+        bytes
+    }
+
+    fn serialize_varint<W: Write, O: Options>(ser: &mut ::ser::Serializer<W, O>, mut n: usize) -> Result<()> {
```
strega-nil: This should take a `u64`, since in theory you could have sizes > the range of usize.

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+
+        while n > 127 {
+            (128 | n as u8).serialize(&mut *ser)?;
+            n >>= 7;
+        }
+
+        (n as u8).serialize(ser)
+    }
+
+    fn deserialize_varint<'de, R: BincodeRead<'de>, O: Options>(de: &mut ::de::Deserializer<R, O>) -> Result<usize> {
```
strega-nil: This should return a `u64`

src/internal.rs
```diff
@@ -189,3 +189,140 @@ impl SizeLimit for Infinite {
...
+
+        bytes
+    }
+
+    fn serialize_varint<W: Write, O: Options>(ser: &mut ::ser::Serializer<W, O>, mut n: usize) -> Result<()> {
+        use serde::Serialize;
+
+        while n > 127 {
+            (128 | n as u8).serialize(&mut *ser)?;
+            n >>= 7;
```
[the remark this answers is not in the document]
ZoeyR: The new CBOR-style varints follow the endian system

fabricedesre:
Hi! Is there anything blocking that to land?

strega-nil:
@fabricedesre COVID, but also I'd like to add varint encoding forall integers.

Zaxar163:
When this PR will be accepted?

ZoeyR:
@Zaxar163 once we've worked out the kinks with signed number varints.

alanjds:
Hi. When will this be released as a crate?

ZoeyR:
Hi @alanjds this was release a few months ago as part of bincode 1.3

PR #310, opened 2020-03-19
expose default options and serializers

ZoeyR:
This is a **_very_** early PR that explores if its possible to expose the `Serializer` and `Deserializer` in a way that would allow us to expand options in the future.

## Basic Design
- expose the `DefaultOptions` and `OptionsExt` traits publicly.
- expose `Serializer` and `Deserializer` publicly.
- keep all other traits and structs `Compound, Options` private

I believe this would allow us to implement any future improvements in a backwards compatible manner. `OptionsExt` can't be implemented by 3rd party crates since it is sealed, so adding new methods should be fine.

## Remarks
- A custom ByteOrder trait was created that provides a thin wrapper over the `byteorder` trait.
  This enables us to add features to our endian behavior in the future without breaking the public API. e.g. making varint respect the endian choices.

## Unanswered Questions
- How should this be integrated with the current `config()` API?
- Who (if anyone) is using the non-public `SerializerAcceptor` trait, and does this make that obsolete?

Tagging @jdm since (I think?) you would have perspective if servo is using these non-public APIs
Tagging @dtolnay since you had significant input on the current config implementation

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,6 +12,7 @@ use self::EndianOption::*;
 use self::LimitOption::*;
 
 ///
+#[derive(Copy, Clone)]
 pub struct DefaultOptions(Infinite);
 
 /// A configuration builder trait whose options Bincode will use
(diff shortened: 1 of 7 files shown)


jdm:
Servo only relies on official public APIs to my knowledge.

ZoeyR:
Great to hear! I realized this makes the unwieldy `config_map` macro obsolete. Previously the match statement would have exploded exponentially, but now we won't need it.

src/config.rs
```diff
@@ -1,64 +1,184 @@
...
-    type Limit: SizeLimit + 'static;
-    type Endian: ByteOrder + 'static;
-
-    fn limit(&mut self) -> &mut Self::Limit;
-}
+/// The default options for bincode serialization/deserialization.
+/// Implements OptionsExt to allow building configuration object for non-default settings.
+///
+/// ### Defaults
+/// By default bincode will use little-endian encoding for mult-byte integers, and will not
```
VictorKoenders, line 17: Why does bincode default to little endian? For all my projects I've used network endian (big endian)

ZoeyR: legacy basically. It was little endian at 1.0 and I don't want to break the byte stream for default configurations.

src/config.rs
```diff
@@ -1,64 +1,184 @@
...
+#[derive(Copy, Clone)]
+pub struct DefaultOptions(Infinite);
 
-pub(crate) trait OptionsExt: Options + Sized {
+/// A configuration builder trait whose options Bincode will use
+/// while serializing and deserializing.
+///
+/// ### Options
+/// Endianness: The endianness with which multi-byte integers will be read/written.  *default: little endian*
+/// Limit: The maximum number of bytes that will be read/written in a bincode serialize/deserialize. *default: unlimited*
```
VictorKoenders, line 27: You need to add a newline in here to break these two sentences in rustdoc

src/config.rs
```diff
@@ -1,64 +1,184 @@
...
+    fn serialize<S: ?Sized + serde::Serialize>(&self, t: &S) -> Result<Vec<u8>> {
+        ::internal::serialize(t, self.clone())
+    }
 
+    /// Returns the size that an object would be if serialized using Bincode with this configuration
     #[inline(always)]
-    fn limit(&mut self) -> &mut Self::Limit {
-        (*self).limit()
+    fn serialized_size<T: ?Sized + serde::Serialize>(&self, t: &T) -> Result<u64> {
+        ::internal::serialized_size(t, self.clone())
```
VictorKoenders, line 74: I think it would be best if all these functions consumed `self` instead of taking a reference and calling `.clone()`. This gives the caller the option to call `.clone()` themselves or let rust simply consume the `Options`.

Currently the types are zero-sized so it won't matter in performance. 

Changing this to consume `self` is more in-line with the other (existing) methods.

src/config.rs
```diff
@@ -1,64 +1,184 @@
...
-    fn limit(&mut self) -> &mut Self::Limit;
-}
+/// The default options for bincode serialization/deserialization.
+/// Implements OptionsExt to allow building configuration object for non-default settings.
+///
+/// ### Defaults
+/// By default bincode will use little-endian encoding for mult-byte integers, and will not
+/// limit the number of serialized/deserialized bytes.
+#[derive(Copy, Clone)]
+pub struct DefaultOptions(Infinite);
```
VictorKoenders, line 20: Nitpick: DefaultOptions is defined here, below this is the `OptionsExt` trait, and below that is `impl DefaultOptions`

src/config.rs
```diff
@@ -69,6 +189,69 @@ impl Options for DefaultOptions {
     }
 }
 
+/// A SizeLimit that restricts serialized or deserialized messages from
+/// exceeding a certain byte length.
+#[derive(Copy, Clone)]
+pub struct Bounded(pub u64);
```
VictorKoenders, line 195: Should probably be named "UpperBound" or "MaxSized"

src/config.rs
```diff
@@ -69,6 +189,69 @@ impl Options for DefaultOptions {
...
+
+    #[inline(always)]
+    fn limit(&self) -> Option<u64> {
+        None
+    }
+}
+
+/// Little-endian byte ordering.
+#[derive(Copy, Clone)]
+pub struct LittleEndian;
```
VictorKoenders, line 233: What is the reason for introducing new types for these endians? It might be better to just re-export the `byteorder` ones and implement the `BincodeByteOrder` trait on those.

VictorKoenders, line 233: Missing NetworkEndian

VictorKoenders: Implementing `impl<T> BincodeByteOrder for T where T: byteorder::ByteOrder { type Endian = T; }` is probably sufficient

src/config.rs
```diff
@@ -333,32 +524,29 @@ impl Config {
...
+    /// A trait for stopping serialization and deserialization when a certain limit has been reached.
+    pub trait SizeLimit: Clone {
+        /// Tells the SizeLimit that a certain number of bytes has been
+        /// read or written.  Returns Err if the limit has been exceeded.
+        fn add(&mut self, n: u64) -> Result<()>;
+        /// Returns the hard limit (if one exists)
+        fn limit(&self) -> Option<u64>;
+    }
+
+    pub trait BincodeByteOrder: Clone {
```
VictorKoenders, line 549: `BincodeByteOrder` is only ever used as a type, not as an instance. This would make the `: Clone` constraint unneeded.

src/config.rs
```diff
@@ -333,32 +524,29 @@ impl Config {
...
-    {
-        config_map!(self, opts => {
-            let mut deserializer = ::de::Deserializer::new(reader, opts);
-            acceptor.accept(&mut deserializer)
-        })
+mod internal {
+    use super::*;
+    use byteorder::ByteOrder;
+
+    pub trait Options: Clone {
```
VictorKoenders, line 533: I'm not 100% sure this `Clone` is needed. It is only passed to `ser::SizeChecker` to check a size limit, and to pre-allocate a Vec. It might be too much effort to remove this constraint though.

src/config.rs
```diff
@@ -333,32 +524,29 @@ impl Config {
...
-    where
-        A: SerializerAcceptor,
-        W: Write,
-    {
-        config_map!(self, opts => {
-            let mut serializer = ::ser::Serializer::new(writer, opts);
-            acceptor.accept(&mut serializer)
-        })
+    /// A trait for stopping serialization and deserialization when a certain limit has been reached.
+    pub trait SizeLimit: Clone {
```
VictorKoenders, line 541: For the record, this Clone is required as it is used by `Bounded`

src/lib.rs
```diff
@@ -79,6 +56,7 @@ pub trait SerializerAcceptor {
 /// |------------|------------|
 /// | Unlimited  | Little     |
 #[inline(always)]
+#[deprecated(since = "1.3.0", note = "please use `DefaultOptions::new()` instead")]
 pub fn config() -> Config {
```
VictorKoenders, line 60: Making this `fn config()` deprecated throws a lot of warnings in the tests. This should be fixed before this PR is merged (but after an API is settled)

src/ser/mod.rs
```diff
@@ -61,49 +61,65 @@ impl<'a, W: Write, O: Options> serde::Serializer for &'a mut Serializer<W, O> {
     }
 
     fn serialize_u16(self, v: u16) -> Result<()> {
-        self.writer.write_u16::<O::Endian>(v).map_err(Into::into)
+        self.writer
+            .write_u16::<<O::Endian as BincodeByteOrder>::Endian>(v)
```
VictorKoenders, line 65: If you wanted to you could be fancy and create newtype
`type GetEndian<O> = <<O as Options>::Endian as BincodeByteOrder>::Endian;`
Then replace all these calls with
`.write_u16::<GetEndian<O>>(v)`

VictorKoenders:
One of the things I dislike is that a caller has to import both `DefaultOptions` and `OptionsExt`, which might cause for confusion. This could be improved in two ways:
- Make defaults for multiple combinations (e.g. `BigEndianNoLimit`, `LittleEndianBounded`)
- Duplicate all methods on `DefaultOptions` so the trait is not always needed (might cause conflicts?)

Note: a lot of crates do this, it seems acceptable to just let this be.

src/config.rs
ZoeyR: I created my own types in case, for some reason, I would need to add data to them in the future. Of course thinking about it, I wouldn't be able to anyways.

The custom BincodeByteOrder trait is necessary though. Since I probably will need to add methods to it at a future date.

src/config.rs
ZoeyR: I actually don't need it. Incoming updates will show how :D

src/config.rs
ZoeyR: I prefer `Bounded` and its been that way for a while so I'll leave it.

src/config.rs
ZoeyR: Since `NetworkEndian` is always `BigEndian` the user can just use `BigEndian`.

src/ser/mod.rs
ZoeyR: I'll leave this alone for now.

Issue #251, opened 2018-10-09
Error building Example code in Readme


ltfschoen:
if I create a new project with
```
cargo new bincode_test;
cargo install cargo-edit;
cargo add bincode serde_derive;
cd bincode_test;
```
And then replace the contents of bincode_test/src/main.rs with the [Example](https://github.com/TyOverby/bincode#example) code from the README.md. 

Then when I try to build the code with:
```
cargo build
```
It gives error:
```
error[E0463]: can't find crate for `_serde`
 --> src/main.rs:7:21
  |
7 | #[derive(Serialize, Deserialize, PartialEq, Debug)]
  |                     ^^^^^^^^^^^ can't find crate
```

System:
cargo 1.29.0 (524a578d7 2018-08-05)
rustc 1.29.1 (b801ae664 2018-09-20)

TyOverby:
`cargo add serde`

jounathaen:
Same problem here. I could fix it by adding
```rust
#[macro_use]
extern crate serde_derive;
```
To the example

robinmoussu:
I had more or less the same issue. I solved it with adding in my Cargo.toml

```toml
[dependencies]
bincode = "1.2"
serde =  { version = "1.0", features = ["derive"] }
```

PR #317, opened 2020-04-12
Add dependencies for Cargo.toml in the example

robinmoussu:
By default the `Derive` macro of Serde isn't available. This cause the example
in the Readme to fail. This commit give one possible way to solve the issue.

This should solve #251

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -27,8 +27,22 @@ library.
 
 ## Example
 
+#### **`Cargo.toml`**
+
+```toml
+[dependencies]
+bincode = "1.2"
+serde =  { version = "1.0", features = ["derive"] }
+```
+
+#### **`main.rs`**
+
 ```rust
-use serde::{Serialize, Deserialize};
+#[macro_use]
+extern crate serde_derive;
+extern crate bincode;
+
+use bincode::{deserialize, serialize};
 
 #[derive(Serialize, Deserialize, PartialEq, Debug)]
 struct Entity {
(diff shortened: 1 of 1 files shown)


robinmoussu:
Note: I don't know witch one is better between the solution I proposed in this PR, or adding the following code to `main.rs`:

```rust
#[macro_use]
extern crate serde_derive;
```

ZoeyR:
I think it would make more sense to

```rust
#[macro_use]
extern crate serde_derive;
```

since this crate still builds in 2015 mode. That way if we added automated testing of our README code snippets later they would compile.

ZoeyR:
I've marked this PR as stale since there hasn't been activity for three months. The PR will be closed in two weeks unless there is new activity.

robinmoussu:
Sorry for the radio silence, I totally forgot this PR.

robinmoussu:
Sorry for the radio silence, I totally forgot this PR.

readme.md
```diff
@@ -27,8 +27,22 @@ library.
 
 ## Example
 
+#### **`Cargo.toml`**
+
+```toml
+[dependencies]
+bincode = "1.2"
+serde =  { version = "1.0", features = ["derive"] }
```
ZoeyR, line 35: Since we're designing this for 2015 edition compilation, this doesn't need the derive feature and a dependency for `serde_derive` needs to be added

robinmoussu: Shouldn't two examples be made? One for consuming it with the 2015 edition and one for consuming it with the 2018 edition?

ZoeyR: `#[macro_use]` is still valid in the 2018 edition. And when we eventually get automated testing of the readme working we would have to make it 2015.

PR #332, opened 2020-06-14
Fix Deserializer::{from_slice, with_reader} types

alyssais:
Having these in the `impl` block with a generic `R` paramter would make them unusable, at least without type annotations:

    error[E0282]: type annotations needed
      --> msg_socket2/src/socket.rs:45:32
       |
    45 |         let mut deserializer = Deserializer::with_reader(bytes.as_slice(), DefaultOptions::new());
       |                                ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `R`

    error: aborting due to previous error

    For more information about this error, try `rustc --explain E0282`.

Moving these into separate `impl` blocks, which set the type of the `Deserializer` to the return type of the functions, fixes this error.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -42,23 +42,27 @@ macro_rules! impl_deserialize_literal {
     };
 }
 
-impl<'de, R: BincodeRead<'de>, O: Options> Deserializer<R, O> {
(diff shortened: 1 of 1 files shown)


ZoeyR:
I'm surprised that it can't infer the type in those cases. Regardless, good catch!

PR #336, opened 2020-06-24
fix logic error in fill_buffer

### LINKED ISSUE #335: Regression in 1.3.0: memory allocation of 8391155392347897856 bytes failed
Can be reproduced by adding the following test to the bottom of src/lib.rs in https://github.com/sharkdp/bat/tree/b09d245dea7e2f45cf461fda978f3256e7162465 and running `cargo update` and `cargo test --lib bincode_repro`. It succeeds with bincode 1.2.1 and aborts with 1.3.0.

 ```rust
#[test]
fn bincode_repro() {
     use syntect::{dumps::from_binary, parsing::SyntaxSet};
     from_binary::<SyntaxSet>(include_bytes!("../assets/syntaxes.bin"));
}
```

```console
memory allocation of 8391155392347897856 bytes failederror: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/git/tmp/bat/target/debug/deps/bat-3446bc567e169f19 repro` (signal: 6, SIGABRT: process abort signal)
```

I've yanked 1.3.0 for now since it's important to keep old serialized data deserializing across minor versions.

ZoeyR:
Closes #335


ZoeyR:
I can design a small unit test that makes sure this function behaves properly. The test suite probably doesn't have complex enough types to trigger this.

Issue #266, opened 2019-05-13
Serialization in other languages


NN-Binary:
Hi,

Anyway you can provide an example to do the ser in Javascript?
Would be awesome for Rust servers.

Thanks man!

TyOverby:
You could compile a Rust program to webassembly and do it there?

Otherwise, I'd recommend using Serde with msgpack or bson

ZoeyR:
I'm turning this into a tracking issue for serialization in other languages. Here are the languages I would like to see:

- [ ] C++
- [ ] Javascript
- [ ] C#
- [ ] Java

Feel free to suggest more

mindv0rtex:
How about Python?

ZoeyR:
I don't have any experience in python, but that would be a good thing to see this in! Once I have a spec written I can also get help from the community to write implementations.

CRRRRR2016:
Hi, I want to have communication between two micro-servers, but one of them is based on rust and the other is based on go.How can I deserialize the messages from the rust server and serialize the messages I have to send to rust server by go ?

Issue #240, opened 2018-07-24
Don't preallocate so much on vectors and strings.


TyOverby:
https://github.com/TyOverby/bincode/issues/239#issuecomment-407527129

I don't think that this much preallocation is necessary.  We should be able to preallocate several megs at a time without hurting performance too much.

PR #346, opened 2020-08-13
Address questions regarding suitability for storage and untrusted inputs

### LINKED ISSUE #216: Readme doesn't mention endian behavior

### LINKED ISSUE #345: Please address suitability for untrusted inputs in docs
A question that I have run into in multiple times on different projects is: How suitable is bincode for network communication?

# Example scenario

A typical scenario is a networking application that sends messages between two participants (peer-to-peer or client server). By default, one can reach for JSON at this point, as the libraries are built against a spec and are safe dealing with hostile inputs (provided they are size-limited).

`bincode` strikes me as something that (at least initially) was intended for trusted communication, possibly in-memory communication, potentially sacrificing support for input from untrusted sources for extra speed.

This question seems to have been asked directly or in a roundabout way in multiple other issues: https://github.com/servo/bincode/issues/216, https://github.com/servo/bincode/issues/136, https://github.com/servo/bincode/issues/221, https://github.com/servo/bincode/issues/240, https://github.com/servo/bincode/issues/255, https://github.com/servo/bincode/issues/266.

## Alternatives

There is a real hole in the Rust ecosystem that yearns to be filled with a compact binary format that can be used to send unversioned, untrusted data with serde support. Certainly [protobuf](https://developers.google.com/protocol-buffers), [captain proto](https://capnproto.org/), [flatbuffers](https://google.github.io/flatbuffers/) exist, but they all tackle a grander scope and come with extra hoops to jump through - at least I just want to slap a `#[derive(Serialize)]` onto my struct and send it more efficiently than JSON across the aether.

Of the available, mature alternatives, msgpack and bson are mentioned, but these are self-describing and thus "waste" space by serializing struct field names as well.

There are other non-established contenders [listed in the serde docs](https://docs.rs/serde/1.0.114/serde/#data-formats), but many fail to address this issue in a precise manner in their docs as well.

# The list

To judge suitability for using `bincode` for a specific project, at least the following questions need to be answered:

1. Is the format stable/well-defined, i.e. does not change across minor versions. This is typically important when storing data using `bincode` long-term.

2. Is the format invariant over differences in byte-order, that is will a big-endian machine produce the same output as a little-endian machine?

3. Is it space-efficient for binary data, i.e. will it be space efficient for large binary data?

4. Is it written with "hostile" data in mind? Rust's memory safety goes a long way of course, but some things like handling invalid inputs or not pre-allocating large amounts of memory when receiving a message are required, and typically a performance-security trade-off.

I may have missed some points, but these came to mind immediately.

It would be great if these could be answered in the docs - I will happily volunteer to write a note and add it to the `README.md` or core module docs, but each of these need an answer from someone who knows the actual source code well.

Each question will actually need two answers, one addressing the status quo and another whether or not this is an accidental or intended property. Any project that relies on bincode certainly needs to know the guarantees it can expect from future versions.

Note that answering "don't know and won't ever guarantee it" on each of these is perfectly fine - I would just like to see it written down, as I find myself in discussions about this with my colleagues fairly often and having sent them to various GitHub issues, it seems that there is a real need for this information =).

If someone could just let me know informally about the state and goals, I will do my best to turn this into a PR updating the docs.

In the meantime, thanks for six years of work that already went into this crate :)

mbr:
Puts some prominent text into the `readme.md` regarding some use cases
that are likely to be common, along with a few hopefully helpful
pointers to avoid footguns.

Closes #345, closes #216, addresses #240, #266.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -80,8 +80,7 @@ some frequently asked questions regarding use of the crate:
 The encoding format is stable across minor revisions, provided the same
 configuration is used. This should ensure that later versions can still read
 data produced by a previous versions of the library if no major version change
-has occured. Beware that in the Rust ecosystem, for a `0.x` version, any change
-in `x` is considered a **major** version change.
+has occured.
 
 Bincode is invariant over byte-order in the default configuration
 (`bincode::options::DefaultOptions`), making an exchange between different
--- a/readme.md
+++ b/readme.md
@@ -90,19 +90,19 @@ metadata like struct field names in the output format and writes long streams of
 binary data without needing any potentially size-increasing encoding.
 
 As a result, Bincode is suitable for storing data. Be aware that it does not
-implement any sort of data versioning scheme oder file headers, as these
+implement any sort of data versioning scheme or file headers, as these
 features are outside the scope of this crate.
 
 ### Is Bincode suitable for untrusted inputs?
 
 Bincode attempts to protect against hostile data. There is a maximum size
-configuration availble (`bincode::config::Bounded`), but not enabled in the
+configuration available (`bincode::config::Bounded`), but not enabled in the
 default configuration. Enabling it causes pre-allocation size to be limited to
 prevent against memory exhaustion attacks.
 
 Deserializing any incoming data will not cause undefined behavior or memory
 issues, assuming that the deserialization code for the struct is safe itself.
 
-Bincode can be used for untrusted inputs in the sense that it will not create an
+Bincode can be used for untrusted inputs in the sense that it will not create a
 security issues in your application, provided the configuration is changed to enable a
 maximum size limit. Malicious inputs will fail upon deserialization.
--- a/readme.md
+++ b/readme.md
@@ -68,3 +68,41 @@ However, there are some implementation details to be aware of:
   `u32` is enough for all practical uses.
 * `str` is encoded as `(u64, &[u8])`, where the `u64` is the number of
   bytes contained in the encoded string.
+
+## Specification
+
+Bincode's format will eventually be codified into a specification, along with
+its configuration options and default configuration. In the meantime, here are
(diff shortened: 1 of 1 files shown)


mbr:
Please let me know if the style is fine; as the original `readme.md` is pleasantly terse. I did opt for putting it into the readme instead of the docs, as people evaluating a format will likely not dive into the docs straight away, but check readmes first.

mbr:
Also, I have no idea why changing the `readme.md` supposedly affects code coverage ¯\\_(ツ)_/¯

readme.md
```diff
@@ -68,3 +68,41 @@ However, there are some implementation details to be aware of:
...
+its configuration options and default configuration. In the meantime, here are
+some frequently asked questions regarding use of the crate:
+
+### Is Bincode suitable for storage?
+
+The encoding format is stable across minor revisions, provided the same
+configuration is used. This should ensure that later versions can still read
+data produced by a previous versions of the library if no major version change
+has occured. Beware that in the Rust ecosystem, for a `0.x` version, any change
+in `x` is considered a **major** version change.
```
ZoeyR, line 84: This last sentence is probably unnecessary since bincode is post 1.0

mbr: Fixed, only took three months :)

mbr:
Three months for a single sentence is not a terribly speedy response, but if this is still good to go, I've submitted the fix. Sorry, I only ran into this after hitting the "mark all read" button on my email account =(

ZoeyR:
Three months to respond to such a fix is also quite slow on my part 😅

ZoeyR (approved):

goral09:
Should I wait another three months to say "thank you"? ;)

marc-casperlabs:
I'll skip the wait and say thanks right now. Much appreciated! :)

PR #355, opened 2020-10-16
Remove dependence on io::Read for deserialization

akhilles:
This contributes to the eventual goal of full `no_std` support. This set of changes should be backwards compatible.


ZoeyR:
Glad to see these changes. We are actually experimenting with no_std support over at https://github.com/bincode-org/bincode-core

akhilles:
> Glad to see these changes. We are actually experimenting with no_std support over at https://github.com/bincode-org/bincode-core

Nice! will try to contribute there, feel free to close this PR.

PR #370, opened 2020-12-23
Add null terminated string option

arlyon:
Hello!

I am working with a protocol that is 99% compatible with the data structures I already use, bar one thing: it uses c-style strings. I know that bincode is technically its own protocol but if we were to support different string formats it would save me 100 lines of custom code...

...So here's 140 lines of custom code that adds a flag similar to the varint flag that allows the user to select between 'vec-style' encoding and 'c-style' encoding. I have marked it as a draft because I wanted to make sure you were open to merging something like this before I finished it.

Thanks! :christmas_tree:

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -36,16 +36,19 @@ pub(crate) use self::endian::BincodeByteOrder;
 pub(crate) use self::int::IntEncoding;
 pub(crate) use self::internal::*;
 pub(crate) use self::limit::SizeLimit;
+pub(crate) use self::str::StrEncoding;
 pub(crate) use self::trailing::TrailingBytes;
 
 pub use self::endian::{BigEndian, LittleEndian, NativeEndian};
 pub use self::int::{FixintEncoding, VarintEncoding};
 pub use self::limit::{Bounded, Infinite};
+pub use self::str::{LenStrEncoding, NullTerminatedStrEncoding};
 pub use self::trailing::{AllowTrailing, RejectTrailing};
 
 mod endian;
 mod int;
 mod limit;
+mod str;
 mod trailing;
 
 /// The default options for bincode serialization/deserialization.
@@ -96,6 +99,7 @@ impl InternalOptions for DefaultOptions {
     type Limit = Infinite;
     type Endian = LittleEndian;
     type IntEncoding = VarintEncoding;
+    type StrEncoding = LenStrEncoding;
     type Trailing = RejectTrailing;
 
     #[inline(always)]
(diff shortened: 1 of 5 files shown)


arlyon:
The main point of conversation is the implementation for de-serializing the 'c-style' strings. The only thing we can do it iterate until a null byte which looks very abnormal compared to the rest of the code in this library!

What are your thoughts?

pavlukivan:
Rust strings can have null bytes, it should be better to handle CString's like that since they can't have null bytes

ItsiAdam:
I want to work with Rust strings, but serialize to C strings. I never want to have CString references in my code and I know what data is being serialized, so this is safe enough for me and a tidy implementation to boot. 
I don't see any reason why it shouldn't be upstreamed, it's not like it's poised to become a default option...

pavlukivan:
At the very least it should panic when serializing a string with a null byte

ZoeyR:
@pavlukivan brings up a good point about Rust strings being able to have null bytes. I'm not a huge fan of adding footguns to the configuration options where the code seems to work for a while and then fails when someone accidentally adds a null byte to a string.

arlyon:
Yes, I suppose you are right that an explicit panic is probably a better option than silently failing. Will rebase and change the behaviour, though there is obviously a performance cost here.

Perhaps this is worth a feature flag? The extra O(n) iteration on each serialization is a cost not everyone would like to bear.

I've removed the draft tag, as I feel there is consensus on this. Let me know your opinions on the feature flag. Cheers!

arlyon:
I'd like to get this merged because it's been open for a while. What is the issue with CI here?

VictorKoenders:
We were waiting on a different PR because we're planning a 2.0 release soon-ish.

Could you fix the merge conflicts? After that we can merge this.

arlyon:
I have rebased onto trunk and fixed any merge issues. And CI passes!

arlyon:
Not stale just awaiting judgement.

TheBestJohn:
is this gated on 2.0 release? My 2 cents is this is an excellent option to add, it fits well with the fixint encoding and is no more of a footgun than it or Endian declaration. If you turn it on, you should know your data. Feature flag could make sense too for those using bincode specifically for performance at high loads but that's better left to the maintainers to decide. Not super familiar with the codebase but if fixint adds the same O(n) I'd say just leave it.

nmccarty:
First, @arlyon, thank you for submitting code for your idea rather than just a feature request issue, the effort is appreciated. 

After some discussion, we have decided to decline this PR, as it violates our previously unformalized design constraint of "no option should arbitrarily limit the values that should be serialized", or, in other words, you should always be able to serialize any valid value of a type, no matter what options are selected.

There is room for disagreement in the general case, but for the purposes of this project we have decided that "I have suddenly found myself with a use case for NUL delimited strings, but I can't use them without having to break all previously stored serialized values or implementing some janky after-the-fact versioning" to be a fundamentally worse type of footgun than endianness.

arlyon:
Fair conclusion, thanks for all your work on bincode!

Issue #348, opened 2020-08-31
Documentation says varint is the default, but it isn't unless you use Options


joshtriplett:
The documentation at https://docs.rs/bincode/1.3.1/bincode/config/trait.Options.html says that the default int encoding is varint. If you use `DefaultOptions`, that's true: `bincode::DefaultOptions::new().serialize_into(...)` will use varint encoding. However, the top-level functions like [`bincode::serialize_into`](https://docs.rs/bincode/1.3.1/bincode/fn.serialize_into.html) say they use "the default configuration", but they appear to use fixint encoding.

pythonesque:
Also (piggybacking on this issue since it's related), if the default configuration *were* to change to varint, it should be a breaking change--while it's true that the interface is still the same, trying to read data serialized using the fixed int encoding (before 1.3) will otherwise be incompatible with the 1.3 release by default, which is a problem for a serialization crate since messages in the old format can be stored in files.

So I suggest performing a major version bump before actually making this the default (I also didn't check whether the Options interface was new to 1.3, but if it was then the upgrade to 1.3 probably should have been a major version update regardless).

ZoeyR:
@pythonesque 1.3 introduced a new options interface, this was not a breaking change since the old interface was left in tact and still works the same way. For a similar reason the basic `bincode::serialize` methods were left with the old configuration. Changing their behavior would be a breaking change and I am committed not to break the crate without a very good reason. The documentation should be clearer though so I will work on updating that.

joshtriplett:
On Fri, Sep 11, 2020 at 10:01:40AM -0700, Zoey Riordan wrote:
> @pythonesque 1.3 introduced a new options interface, this was not a breaking change since the old interface was left in tact and still works the same way. For a similar reason the basic `bincode::serialize` methods were left with the old configuration. Changing their behavior would be a breaking change and I am committed not to break the crate without a very good reason. The documentation should be clearer though so I will work on updating that.

Updating the documentation would suffice. The top-level functions should
state what configuration they use and that it isn't the default, and
make it clear that you should create a `bincode::DefaultOptions::new()`
and use its methods if you want the defaults.

PR #373, opened 2021-02-22
Update docs to highlight differences between DefaultOptions and fns

apgoetz:
Due to historical reasons, the default options used by bincode for the helper functions at the root of the crate differ from the options in the `config` module.

Changing the default options is a breaking change, so until a decision is made, we can at least document the current behavior.

Related issue: #348


ZoeyR:
Looks great! thanks.

PR #375, opened 2021-02-23
clarify msrv support

ghost:


VictorKoenders (approved):

VictorKoenders (approved):

dtolnay (commented):
This PR breaks downstream projects and is not the right fix I think. See https://github.com/bincode-org/bincode/issues/388.

PR #402, opened 2021-08-06
Unignore doctests where possible

### LINKED ISSUE #363: Example for Deserializer uses `bytes_read` method which doesn't exist
The sample code for bincode::Deserializer uses a `bytes_read`, which doesn't exist and appears to have been removed years ago.

rukai:
closes https://github.com/bincode-org/bincode/issues/363

Possibly we would be better off just deleting the examples from Serializer and Deserializer?
I dont mind either way.

I left the zigzag example as ignore because its a pseudocode demonstration and getting it compiling would obscure the intent.

### DIFF
--- a/src/config/int.rs
+++ b/src/config/int.rs
@@ -147,7 +147,7 @@ pub struct FixintEncoding;
 /// fn zigzag(v: Signed) -> Unsigned {
 ///     match v {
 ///         0 => 0,
-///         v if v < 0 => |v| * 2 - 1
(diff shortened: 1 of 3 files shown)


rukai:
alright, I rebased it.

ghost:
Sorry I missed this, I'll mark it as not stale for now

VictorKoenders:
Because we're in the middle of a big rewrite for bincode 2, I've taken the liberty to implement your change in a different PR: #421 

Thank you for your contribution!

PR #404, opened 2021-09-01
Fix crates.io badge and link to GitHub Actions page from badge

atouchet:


VictorKoenders:
I have applied this change in #417. We're in the middle of a big rewrite of bincode for the 2.0 release, and I did not want to bother you with updating your branch for that.

Thanks for the contribution!

PR #405, opened 2021-09-10
Exposes in-place deserialization from buffers using SliceReader

dequbed:
Updating values in place is rare, but it is safe to do from a slice.
Adding a function `deserialize_in_place_buffer` allows users to do this
in-place update without having to vendor SliceReader.

Pretty much my situation. I have a mmap-based database from which I want to pull updates into existing structs.

The alternatives are allowing construction of a `SliceReader` from an arbitrary byteslice (i.e. making `SliceReader::new()` `pub` instead of `pub(crate)`) and vendoring the entire code of `SliceReader` verbatim, the latter being what I'm trying to avoid.

EDIT: I just now realized that I could also just create a `bincode::de::Deserializer` and use that with `serde::Deserialize::deserialize_in_place`. However that works less well with my current code structure, so if you feel like this PR doesn't clutter your export table unduly I would still prefer to do it this way.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -208,6 +208,17 @@ pub trait Options: InternalOptions + Sized {
         crate::internal::deserialize_in_place(reader, self, place)
     }
 
+    /// TODO: document
+    #[doc(hidden)]
+    #[inline(always)]
+    fn deserialize_in_place_buffer<'a, T: serde::de::Deserialize<'a>>(
+        self,
+        bytes: &'a [u8],
+        place: &mut T
+    ) -> Result<()> {
+        crate::internal::deserialize_in_place_buffer(bytes, self, place)
+    }
+
     /// Deserializes a slice of bytes with state `seed` using this configuration.
     #[inline(always)]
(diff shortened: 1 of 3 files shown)


VictorKoenders:
Just a heads up: we're currently hard at work on releasing bincode 2.0. In bincode 2.0 we currently do not have a "deserialize_in_place" method, although we might want to add that in the future.

Because of this I do not expect to merge this PR any time soon.

dequbed:
> Just a heads up: we're currently hard at work on releasing bincode 2.0. In bincode 2.0 we currently do not have a "deserialize_in_place" method, although we might want to add that in the future.
> 
> Because of this I do not expect to merge this PR any time soon.

Ah, thanks for the heads up. In that case I'll close this PR since I can solve my issue in a different way and this PR would go stale with any v2.0 and just add to your upgrading workload if merged now.

If you or somebody else wants to implement this at some later date feel free to re-open or just take the patches as basis, they are licensed under the same license that the current and any future versions of bincode are under.

PR #408, opened 2021-10-07
Fix CI on trunk

ghost:
CI was currently broken for two reasons
- `trunk` was not compliant with the 1.41 MSRV check
- `clippy` had new warnings

Since bincode 2.0 is not even in alpha yet MSRV should not be locked in place. This PR removes the MSRV checks for now. Additionally the few clippy warnings have been resolved.


.github/workflows/rust.yml
```diff
@@ -25,7 +25,7 @@
             "stable",
             "beta",
             "nightly",
-            "1.41.0"
+            # "1.41.0"
```
VictorKoenders: The deserde branch has this line:
```json
# "1.55.0" TODO: Pick latest stable version when we release 2.
```
Do you think we should copy over the TODO?

VictorKoenders (approved):

.github/workflows/rust.yml
ghost: Hmm yeah, I'll update

PR #409, opened 2021-10-07
Make bincode_derive 0 dependencies

VictorKoenders:
This PR removes `syn`, `proc_macro2` and `quote` from `bincode_derive`.

We still have to keep `proc_macro2` as a dev dependency as it is otherwise impossible to test the parser. This is automatically used instead of `proc_macro` when running tests.

The change is roughly split into 2 parts:

# Parsing

Most of parsing is done in the `derive/src/parse/` folder. This will generate the following types:
- `Visibility`, not being used currently
- `DataType` either `Struct` or `Enum`, with the name of the data type being parsed
- `Generics` the generics part of the type, e.g. `struct Foo<'a>`
- `GenericConstraints` the "where" part of the type

# Generate

Generating the code implementation is done in either `derive/src/derive_enum.rs` and `derive/src/derive_struct.rs`.

This is supported by the structs in `derive/src/generate`. The most notable points of this module are:
- `StreamBuilder` is a thin but friendly wrapper around `TokenStream`
- `Generator` is the base type of the code generator. This has helper methods to generate implementations:
  - `ImplFor` is a helper struct for a single `impl A for B` construction. In this functions can be defined:
    - `GenerateFnBody` is a helper struct for a single function in the above `impl`. This is created with a callback to `FnBuilder` which helps set some properties. `GenerateFnBody` has a `stream()` function which returns ` StreamBuilder` for the function.

The code is tested against the current cases in `tests/derive.rs`, **however** we'll need a lot more test cases before this PR can be merged.

For testing purposes, all generated code is outputted to the current `target` folder, under file name `<struct/enum name>_Encodeable.rs` and `<struct/enum name>_Decodeable.rs`. This can help with debugging.


Cargo.toml
```diff
@@ -21,6 +21,7 @@ license = "MIT"
 description = "A binary serialization / deserialization strategy for transforming structs into bytes and vice versa!"
 
 edition = "2018"
+resolver = "2"
```
ghost: Did cargo do this automatically?

derive/src/parse/body.rs
```diff
@@ -0,0 +1,313 @@
...
+    pub fn is_struct_variant(&self) -> bool {
+        // An enum variant is a struct variant if it has any fields with a name
+        // enum Foo { Bar { a: u32, b: u32 } }
+        self.fields
+            .as_ref()
+            .map(|f| f.iter().any(|f| f.ident.is_some()))
+            .unwrap_or(false)
+    }
+    pub fn is_tuple_variant(&self) -> bool {
+        // An enum variant is a struct variant if it has no fields with a name
```
ghost: This comment should say `tuple variant`

derive/src/derive_enum.rs
```diff
@@ -1,242 +1,191 @@
...
+        fn_body.group(Delimiter::Brace, |match_body| {
+            for (variant_index, variant) in variants.into_iter().enumerate() {
+                // Self::Variant
+                match_body.ident_str("Self");
+                match_body.puncts("::");
+                match_body.ident(variant.name.clone());
+
+                // if we have any fields, declare them here
+                if let Some(fields) = variant.fields.as_ref() {
+                    let delimiter = if variant.is_struct_variant() {
```
ghost: I think it would make more sense to have a `variant.kind()` method that returns an enum of `Struct, Tuple, Unit` instead of having separate `is` methods.

derive/src/error.rs
```diff
@@ -1,24 +1,98 @@
...
-    UnionNotSupported,
+    UnknownVisibility(Span),
+    UnknownDataType(Span),
+    InvalidRustSyntax(Span),
+    ExpectedIdent(Span),
+    // UnionNotSupported,
+}
+
+impl PartialEq for Error {
+    fn eq(&self, other: &Error) -> bool {
```
ghost: Why are the `Span`s not compared for equality?

derive/src/derive_struct.rs
```diff
@@ -1,145 +1,106 @@
...
+                .with_return_type("Result<(), bincode::error::EncodeError>")
+        });
+        let fn_body = fn_body.stream();
+        for (idx, field) in fields.iter().enumerate() {
+            let field_name = field
+                .ident
+                .as_ref()
+                .map(|idx| idx.to_string())
+                .unwrap_or_else(|| idx.to_string());
+            fn_body.push_parsed(format!(
```
ghost: It would be useful to somehow add the span of the original field to this part of the token stream. That way if the field doesn't implement `Encodable` the compiler error will produce lines at the correct place. This probably applies other places, but I'll just comment it here for now.

derive/src/generate/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+        generator.stream.append(builder);
+
+        let group = StreamBuilder::new();
+        Self { generator, group }
+    }
+
+    pub fn generate_fn<'b>(
+        &'b mut self,
+        name: &str,
+        builder: impl FnOnce(FnBuilder) -> FnBuilder,
```
ghost: I'm not a huge fan of the closure API. Would be possible to change this to return a builder with a `build()` method instead?

derive/src/generate/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+        Self {
+            name: name.into(),
+            lifetime_and_generics: Vec::new(),
+            self_arg: FnSelfArg::None,
+            args: Vec::new(),
+            return_type: None,
+        }
+    }
+
+    // pub fn with_lifetime(
```
ghost: Remove commented out code please

derive/src/parse/body.rs
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+        let group = assume_group(input.next());
+        let mut stream = group.stream().into_iter().peekable();
+        let fields = match group.delimiter() {
+            Delimiter::Brace => Field::parse_named(&mut stream)?,
+            Delimiter::Parenthesis => Field::parse_unnamed(&mut stream)?,
+            _ => return Err(Error::InvalidRustSyntax(group.span())),
+        };
+        assert!(
```
ghost: Could this be turned into a debug_assert or an Error variant instead?

derive/src/parse/visibility.rs
```diff
@@ -0,0 +1,80 @@
...
+                // Consume this token
+                let ident = super::assume_ident(input.next());
+
+                // check if the next token is `pub(crate)`
+                if let Some(TokenTree::Group(_)) = input.peek() {
+                    let group = super::assume_group(input.next());
+                    let mut group_stream = group.stream().into_iter();
+                    return match (group_stream.next(), group_stream.next()) {
+                        (Some(TokenTree::Ident(ident)), None) => {
+                            if super::ident_eq(&ident, "crate") {
```
ghost: other visibilities than `crate` are allowed

ghost (changes requested):

Cargo.toml
VictorKoenders: No I added this manually. Resolver 2 is standard in the 2021 edition, and I think this won't actually do anything if it's not in the binary (plus we have almost no dependencies). I'll remove it

derive/src/parse/body.rs
VictorKoenders: Reworked the struct/enum variant fields

derive/src/derive_enum.rs
VictorKoenders: Done!

derive/src/error.rs
VictorKoenders: `PartialEq` was only implemented for `Error` for the tests so that `partial_eq!()` could be used. I've removed the `PartialEq` implementation and used all the `.is_` methods and `.unwrap()`

derive/src/parse/body.rs
VictorKoenders: The `assert` should only panic if the fields of the structs are improperly parsed. I'm assuming the parser is properly tested at this point and we don't need this assert, so I'll remove it.

derive/src/derive_struct.rs
VictorKoenders: Because this is formatted into a string and then passed to `.push_parsed`, we're dependent on the `Span` value of the implementation of the `FromStr` in either `proc_macro` (for users) or `proc_macro2` (during tests).

I'll add a `Span` location to `IdentOrIndex::Index` but I doubt it'll ever come up

derive/src/parse/visibility.rs
VictorKoenders: Because we're not actually using the value of `Visibility` I've changed it to simply consume everything the parenthesis. Visibility is now either `::Default` for no visibility and `::Pub` for `pub` and `pub(anything)`

derive/src/generate/mod.rs
VictorKoenders: I have flipped it around. Configuring the function definition is now a builder pattern, and the body is inside of a closure. The body is a nested group so it makes sense to be a similar closure like `StreamBuilder::group`.

derive/src/error.rs
```diff
@@ -1,24 +1,76 @@
...
 }
 
 impl Error {
     pub fn into_token_stream(self) -> TokenStream {
-        self.into_token_stream_with_span(Span::call_site())
+        let maybe_span = match &self {
+            Error::UnknownDataType(span)
+            | Error::ExpectedIdent(span)
+            | Error::InvalidRustSyntax(span) => Some(*span),
+            // Error::UnionNotSupported => None,
```
ghost: Remove this

src/de/mod.rs
```diff
@@ -6,7 +6,8 @@ mod impls;
 pub mod read;
 pub use self::decoder::Decoder;
 
-pub trait Decodable: for<'de> BorrowDecodable<'de> {
+pub trait Decodable: Sized {
+    // }: for<'de> BorrowDecodable<'de> {
```
ghost: why is this here?

ghost:
Looks like this parser can't handle field attributes (i.e. `#[serde(skip)]`). It shouldn't be too bad to add them because as far as I can tell they can all be ignored.

derive/src/error.rs
VictorKoenders: Removed

src/de/mod.rs
VictorKoenders: Leftover from when I was trying to make the derive crate work, reverted this change

src/de/mod.rs
```diff
@@ -6,7 +6,7 @@ mod impls;
 pub mod read;
 pub use self::decoder::Decoder;
 
-pub trait Decodable: for<'de> BorrowDecodable<'de> {
```
ghost, line 9: I really hate to nitpick but could you remove the `Sized`? Its implied by BorrowDecodable

ghost (approved):

src/de/mod.rs
VictorKoenders: Done

PR #412, opened 2021-10-17
Config rewrite

ghost:


src/config.rs
```diff
@@ -2,11 +2,11 @@
 //!
 //! *Important* make sure you use the same config for encoding and decoding, or else bincode will not work properly.
 //!
-//! To use a config, first create a type of [struct@Default]. This type will implement trait [Config] for further configuration.
+//! To use a config, first create a type of [struct@Configuration]. This type will implement trait [Config] for use with bincode.
```
VictorKoenders: `[struct@Configuration]` could be `[Configuration]`, this used to be there because `Default` would conflict with the trait

src/config.rs
```diff
@@ -38,15 +39,41 @@ use core::marker::PhantomData;
...
+    _a: PhantomData<A>,
+}
+
+#[allow(clippy::new_without_default)]
+impl Configuration {
+    /// The default config. By default this will be:
+    /// - Little endian
+    /// - Variable int encoding
+    /// - Skip fixed array length
+    pub fn new() -> Self {
```
VictorKoenders: This should probably be renamed to `standard` or something, we can bikeshed it in the future but I want to be clear it's not `new`

VictorKoenders: Do we know what the `legacy` (bincode 1) config was? Maybe we can add that immediately?

src/config.rs
```diff
@@ -38,15 +39,41 @@ use core::marker::PhantomData;
...
+            _i: PhantomData,
+            _a: PhantomData,
+        }
+    }
+
     /// Makes bincode encode all integer types in big endian.
-    fn with_big_endian(self) -> BigEndian<Self> {
-        BigEndian { _pd: PhantomData }
+    pub fn with_big_endian(self) -> Configuration<BigEndian, I, A> {
+        Self::generate()
```
VictorKoenders, line 78: Nitpick: `generate` could be a local fn instead of an impl fn, which would save a couple of keystrokes

src/config.rs
```diff
@@ -38,15 +39,41 @@ use core::marker::PhantomData;
...
+    /// - Little endian
+    /// - Variable int encoding
+    /// - Skip fixed array length
+    pub fn new() -> Self {
+        Self::generate()
+    }
+}
+
+impl<E, I, A> Configuration<E, I, A> {
+    fn generate<_E, _I, _A>() -> Configuration<_E, _I, _A> {
```
VictorKoenders, line 68: The current implementation means we can't transfer state between the previous instance of `Configuration` and the next one. For now I can only think of config options that are totally encoded in the type system, do you foresee any configs that could possibly contain variables that would need transferring? Maybe callback functions or something?

src/config.rs
VictorKoenders: Discussed with Zoey and we decided to keep `generate` as it is

src/config.rs
VictorKoenders: Discussed with Zoey and we decided to leave it like this until it becomes a problem. It's internal so at that point we can figure it out

VictorKoenders (approved):

PR #413, opened 2021-10-17
Feature/deserde

ghost:


Cargo.toml
```diff
@@ -1,35 +1,54 @@
...
-codegen-units = 1
-debug = 1
+[workspace]
+members = [
+    "derive"
+]
+
+[package]
+name = "bincode"
+version = "2.0.0-dev" # remember to update html_root_url and bincode_derive
```
VictorKoenders: Should we rename this to `-alpha.0` before merging into trunk? Semver pre-release versions are alphabetically ordered  which means `-dev` is after `-alpha`.

Alternatively we can call it `-aaaa` to make sure it's always older

Cargo.toml
```diff
@@ -1,35 +1,54 @@
...
+std = ["alloc"]
+alloc = []
+atomic = []
+derive = ["bincode_derive"]
+
+[dependencies]
+bincode_derive = { path = "derive", version = "2.0.0-dev", optional = true }
+serde = { version = "1.0.130", optional = true }
+
+# Used for derive tests
```
VictorKoenders: Should be `# Used for tests`

derive/src/error.rs
```diff
@@ -0,0 +1,70 @@
...
+            Error::UnknownDataType(span)
+            | Error::ExpectedIdent(span)
+            | Error::InvalidRustSyntax(span) => Some(*span),
+        };
+        self.throw_with_span(maybe_span.unwrap_or_else(Span::call_site))
+    }
+
+    pub fn throw_with_span(self, span: Span) -> TokenStream {
+        // compile_error!($message)
+        vec![
```
VictorKoenders: TODO (me): we can use the generator API for this

derive/src/generate/generate_fn.rs
```diff
@@ -0,0 +1,134 @@
+use super::{ImplFor, StreamBuilder};
+use crate::prelude::Delimiter;
+pub struct FnBuilder<'a, 'b> {
```
VictorKoenders, line 5: Nitpick: newline between `use ...` and `struct ...`

derive/src/generate/generator.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[must_use]
+pub struct Generator {
+    pub(super) name: Ident,
+    pub(super) generics: Option<Generics>,
+    pub(super) generic_constraints: Option<GenericConstraints>,
+    pub(super) stream: StreamBuilder,
+}
+
+impl Generator {
```
VictorKoenders, line 14: TODO (me): I think these functions need to be documented, especially because I'm considering making this a seperate crate, but I won't block this PR on it

derive/src/generate/stream_builder.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    pub fn push(&mut self, item: impl Into<TokenTree>) {
+        self.stream.extend([item.into()]);
+    }
+
+    pub fn push_parsed(&mut self, item: impl AsRef<str>) {
+        self.stream
+            .extend(TokenStream::from_str(item.as_ref()).unwrap_or_else(|e| {
+                panic!(
```
VictorKoenders: TODO (me): this should probably return a `Result`

docs/spec.md
```diff
@@ -0,0 +1,160 @@
+# Serialization specification
```
VictorKoenders, line 1: @ZoeyR could you review this document to see if I missed something?

docs/spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Fixed size integers are encoded directly
+- Enum discriminants are encoded as u32
+- Lengths and usize are encoded as u64
+
+See the documentation of [FixintEncoding](https://docs.rs/bincode/latest/bincode/config/struct.FixintEncoding.html) for more information.
+
+## Enums
+
+Enums are encoded with their variant first, followed by optionally the variant fields. The variant index is based on the `IntEncoding` during serilization.
```
VictorKoenders, line 57: This should explicitly mention that it's serialized as a `u32`

docs/spec.md
```diff
@@ -0,0 +1,160 @@
...
+Boolean types are encoded with 1 byte for each boolean type, with `0` being `false`, `1` being true. Whilst deserilizing every other value will throw an error.
+
+All basic numeric types will be encoded based on the configured [IntEncoding](#IntEncoding).
+
+All floating point types will take up exactly 4 (for `f32`) or 8 (for `f64`) bytes.
+
+All tuples have no additional bytes, and are encoded in their specified order, e.g.
+```rs
+let tuple = (u32::min_value(), i32::max_value()); // 8 bytes
+let encoded = bincode::encode_to_vec_with_options(&tuple, Options::default().with_fixint_encoding()).unwrap();
```
VictorKoenders: The examples in this document should match the new `Config` api.

VictorKoenders: Also I think we can include this document in rustdoc with `#[doc = include_str!("../docs/spec.md")]` somewhere in the source code. This will help with testing the examples and it'll show up on docs.rs

docs/spec.md
```diff
@@ -0,0 +1,160 @@
...
+    0, 0, 0, 2, // first variant, C
+    0, 0, 0, 0  // C has 1 named field which is a u32, so 4 bytes
+]);
+```
+
+# Collections
+
+Collections are encoded with their length value first, following by each entry of the collection. The length value is based on your `IntEncoding`.
+
+**note**: fixed array length do not have their `len` encoded. See [Arrays](#arrays)
```
VictorKoenders, line 97: This is dependent on a config flag now

docs/spec.md
```diff
@@ -0,0 +1,160 @@
...
+let encoded = bincode::encode_to_vec_with_options(&list, Options::default().with_fixint_encoding()).unwrap();
+assert_eq!(encoded.as_slice(), &[
+    0, 0, 0, 0, 0, 0, 0, 5, // length of the string, 5 bytes
+    b'H', b'e', b'l', b'l', b'o'
+]);
+```
+
+# Arrays
+
+Arrays are encoded *without* a length.
```
VictorKoenders: Same here

src/de/decoder.rs
```diff
@@ -0,0 +1,59 @@
...
+/// ```
+/// # let slice: &[u8] = &[0, 0, 0, 0];
+/// # let some_reader = bincode::de::read::SliceReader::new(slice);
+/// use bincode::de::{DecoderImpl, Decode};
+/// use bincode::config;
+/// let mut decoder = DecoderImpl::new(some_reader, config::Configuration::standard());
+/// // this u32 can be any Decode
+/// let value = u32::decode(&mut decoder).unwrap();
+/// ```
+pub struct DecoderImpl<R, C: Config> {
```
VictorKoenders, line 24: We were considering renaming this to `DecoderState`, as well as adding the number of bytes written which we'll need for the `byte limit` config setting

src/de/impls.rs
```diff
@@ -0,0 +1,547 @@
...
+        decoder.reader().read(&mut array[1..width])?;
+        let res = core::str::from_utf8(&array[..width])
+            .ok()
+            .and_then(|s| s.chars().next())
+            .ok_or(DecodeError::InvalidCharEncoding(array))?;
+        Ok(res)
+    }
+}
+
+impl<'a, 'de: 'a> BorrowDecode<'de> for &'a [u8] {
```
VictorKoenders, line 379: Could `'a` and `'de` be combined into a single lifetime here?

src/de/read.rs
```diff
@@ -1,327 +1,71 @@
...
 impl<'storage> SliceReader<'storage> {
     /// Constructs a slice reader
-    pub(crate) fn new(bytes: &'storage [u8]) -> SliceReader<'storage> {
+    pub fn new(bytes: &'storage [u8]) -> SliceReader<'storage> {
         SliceReader { slice: bytes }
     }
 
     #[inline(always)]
-    fn get_byte_slice(&mut self, length: usize) -> Result<&'storage [u8]> {
+    fn get_byte_slice(&mut self, length: usize) -> Result<&'storage [u8], DecodeError> {
```
VictorKoenders: This function is only used for `BorrowReader::take_bytes` and nowhere else, what was the reason we split this again?

src/enc/encoder.rs
```diff
@@ -0,0 +1,56 @@
+//! Contains
```
VictorKoenders: This documentation should be updated

src/enc/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+mod impls;
+
+use crate::{config::Config, error::EncodeError};
+
+pub mod write;
+
+pub use self::encoder::EncoderImpl;
+use self::write::Writer;
+
+/// Any source that can encode types. This type is most notably implemented for [Encoder].
```
VictorKoenders: This documentation should be updated. This trait is implemented for every type that needs to be encodable.

src/enc/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn writer(&mut self) -> &mut Self::W {
+        T::writer(self)
+    }
+
+    fn config(&self) -> &Self::C {
+        T::config(self)
+    }
+}
+
+pub(crate) mod sealed {
```
VictorKoenders: This `sealed` mod should probably be moved to a central utils file, we use Sealed in more places

src/features/impl_std.rs
```diff
@@ -0,0 +1,324 @@
...
+        match self.read_exact(bytes) {
+            Ok(_) => Ok(()),
+            Err(_) => Err(DecodeError::UnexpectedEnd),
+        }
+    }
+}
+
+/// Encode the given value into any type that implements `std::io::Write`, e.g. `std::fs::File`.
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
+pub fn encode_into_write<E: Encode, W: std::io::Write>(
```
VictorKoenders: I'm thinking we should remove all `encode_into_xxx_with_config` and `decode_from_xxx_with_config` and always make `Config` a mandatory selection by the user.

We'll want to update `Config` in the future for more effective encoding/decoding options, if we provide people with a default in the form of `encode_into_write` which uses `Configuration::standard()` on bincode 2.0 launch, we can never change the default to anything else than `::standard()`. Making the user specify the `Configuration` explicitly will make them more aware that there are new config options available in the future.

VictorKoenders: In addition it'll halve the amount of encode/decode options we have, which will be important because e.g. the `serde` feature will probably double the amount of encode/decode functions (one for `&mut [u8]`, one for `std::io::Write`, etc)

src/features/impl_std.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+}
+
+impl<T> Encode for RwLock<T>
+where
+    T: Encode,
+{
+    fn encode<E: Encoder>(&self, encoder: E) -> Result<(), EncodeError> {
+        let t = self.read().map_err(|_| EncodeError::LockFailed {
+            type_name: core::any::type_name::<Mutex<T>>(),
```
VictorKoenders: TODO(me): Copy paste error, should be `RwLock`

src/features/impl_std.rs
```diff
@@ -0,0 +1,324 @@
...
+        let duration = Duration::decode(decoder)?;
+        Ok(SystemTime::UNIX_EPOCH + duration)
+    }
+}
+
+impl Encode for &'_ Path {
+    fn encode<E: Encoder>(&self, encoder: E) -> Result<(), EncodeError> {
+        match self.to_str() {
+            Some(str) => str.encode(encoder),
+            None => Err(EncodeError::Other("Path contains invalid UTF-8 characters")),
```
VictorKoenders: I introduced the `EncodeError::Other` variant here because I didn't know how many variants we'd have of weird corner cases. I think we can get rid of `Other` for now and make it specifically an `InvalidPathCharacters` error

src/lib.rs
```diff
@@ -1,131 +1,93 @@
...
-pub use crate::de::read::BincodeRead;
-pub use crate::de::Deserializer;
-pub use crate::error::{Error, ErrorKind, Result};
-pub use crate::ser::Serializer;
-
-/// Get a default configuration object.
+/// Encode the given value into the given slice. Returns the amount of bytes that have been written.
 ///
-/// ### Default Configuration:
+/// Will take the [Default] configuration. See the [config] module for more information.
```
VictorKoenders: This is now renamed to `Configuration::standard()`. I'm surprised the CI did not catch this

src/features/impl_std.rs
ghost: I think we still want to keep an "Other" variant. 3rd parties may implement `Encode` and `Decode` for types that have their own correctness rules.

src/features/impl_std.rs
ghost: Wouldn't the serde solution be to just implement `Encode` and `Decode` on any type that implements `Serialize` and `Deserialize`? Then we wouldn't need an extra entry point.

src/de/impls.rs
ghost: That would actually make the implementation more specific than what we want. If the lifetime was `&'de [u8]` then you could only decode a slice that lived as long as the backing storage. You would not be able to decode a slice that lived for a shorter time, which will often be the case.

src/de/read.rs
ghost: I'm not sure, but it might be needed for the `peek_read` optimization that got added in trunk. I'll evaluate.

docs/spec.md
VictorKoenders: Currently we do doctests in a dedicated `specs` module, however because the doctests require both `alloc` and `derive`, we disable the `specs` module when those features are not enabled.

It would be nice if we can ignore the doctests if those features are disabled but still compile the documentation for the user, however I don't know how to do that

src/features/impl_std.rs
VictorKoenders: I'm not sure if that's possible. If it is, then there won't be that issue. However I still think we don't need dedicated `_with_config` methods

src/features/impl_std.rs
VictorKoenders: Added the new variant but kept `Other`

VictorKoenders (approved):

PR #414, opened 2021-10-22
Reintroduce varint optimizations

ghost:


src/de/impls.rs
```diff
@@ -31,10 +31,17 @@ impl Decode for bool {
 }
 
 impl Decode for u8 {
+    #[inline]
     fn decode<D: Decoder>(mut decoder: D) -> Result<Self, DecodeError> {
-        let mut bytes = [0u8; 1];
-        decoder.reader().read(&mut bytes)?;
-        Ok(bytes[0])
+        if let Some(buf) = decoder.reader().peek_read(1) {
```
VictorKoenders, line 36: I'm wondering if this implementation is actually faster, because it always tries to read exactly 1 byte

src/features/impl_std.rs
```diff
@@ -19,30 +20,60 @@ use std::{
 ///
 /// [config]: config/index.html
 #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
-pub fn decode_from_reader<D: Decode, C: Config, R: std::io::Read>(
+pub fn decode_from_std_reader<D: Decode, C: Config, R: std::io::Read>(
```
VictorKoenders: Bikeshed: I think this could either be `_from_reader` (any reader) or `_from_std_read` (for `std::io::Read`)

src/features/impl_std.rs
```diff
@@ -19,30 +20,60 @@ use std::{
...
     fn read(&mut self, bytes: &mut [u8]) -> Result<(), DecodeError> {
         match self.read_exact(bytes) {
             Ok(_) => Ok(()),
             Err(_) => Err(DecodeError::UnexpectedEnd),
         }
     }
+
+    #[inline]
+    fn peek_read(&self, n: usize) -> Option<&[u8]> {
+        self.buffer().get(..n)
```
VictorKoenders, line 62: For this we can do https://doc.rust-lang.org/stable/std/io/trait.BufRead.html#tymethod.fill_buf, which might seem slightly better if the reader is empty at the time

src/varint/decode_unsigned.rs
```diff
@@ -1,42 +1,47 @@
...
+    read.read(&mut bytes)?;
+    match bytes[0] {
+        byte @ 0..=SINGLE_BYTE_MAX => Ok(byte as u16),
         U16_BYTE => {
             let mut bytes = [0u8; 2];
             read.read(&mut bytes)?;
             Ok(match endian {
-                Endian::Big => u16::from_be_bytes(bytes),
-                Endian::Little => u16::from_le_bytes(bytes),
+                Endian::Big => u16::from_be_bytes(bytes) as u16,
```
VictorKoenders: This `as u16` seems unneeded

src/de/impls.rs
ghost: `peek_read` tends to be constant, so if it succeeds then it doesn't have to copy. If it fails it will pretty much always fail.

src/varint/decode_unsigned.rs
```diff
@@ -49,26 +54,26 @@ pub fn varint_decode_u32<R: Reader>(read: &mut R, endian: Endian) -> Result<u32,
...
-        U64_BYTE => Err(DecodeError::InvalidIntegerType {
-            expected: IntegerType::U32,
-            found: IntegerType::U64,
-        }),
-        U128_BYTE => Err(DecodeError::InvalidIntegerType {
-            expected: IntegerType::U32,
-            found: IntegerType::U128,
-        }),
-        x => Ok(x as u32),
+        U64_BYTE => invalid_varint_discriminant(IntegerType::U32, IntegerType::U64),
```
VictorKoenders, line 61: This is missing the `U32_BYTE`

VictorKoenders: Never mind github just collapsed in the middle of hte function

src/features/impl_std.rs
ghost: Isn't that what I did? I added `from_reader` methods to lib that can take any `Reader`

VictorKoenders: This function is named `_from_std_reader`, I'd suggest removing the `er` at the end

src/features/impl_std.rs
VictorKoenders: Better because the documentation of `buffer` is:
> Returns a reference to the internally buffered data.
> Unlike fill_buf, this will not attempt to fill the buffer if it is empty.

ghost: It actually might be worse performance wise because filling the internal buffer would trigger a copy. `peek_read` should ideally not trigger any copies.

src/de/impls.rs
VictorKoenders: Confirmed in a benchmark, is indeed faster with bufreader, not sure why

image

VictorKoenders (approved):

PR #417, opened 2021-10-25
Updated readme.md and added a test for the examples

VictorKoenders:


tests/readme.rs
```diff
@@ -0,0 +1,32 @@
+#![cfg(all(feature = "derive", feature = "alloc"))]
```
ghost: This will be difficult to keep in sync, I recommend adding the readme as a module similar to the spec. Annotate it with `#[cfg(all(feature = "alloc", feature = "derive", doctest))]` so that it only runs during doctests

VictorKoenders: Good idea, done

ghost (approved):

PR #419, opened 2021-10-25
Replace test-all-features with a manual CI matrix

VictorKoenders:
Because the `cargo test-all-features` run was taking almost 10 minutes, I've replaced it by a manual build matrix.

Additionally I've moved the beta and nightly rust versions to the `cargo check`. We probably won't need to run tests on those channels (if we do, we can change it back in the future).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,7 +4,6 @@
     "push": {
       "branches": [
         "trunk",
-        "bincode2/ci_matrix", # Temporary, don't merge this
         "v*.x"
       ]
     },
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -4,7 +4,7 @@
     "push": {
       "branches": [
         "trunk",
-        "feature/deserde*", # Temporary while we work on deserde
+        "bincode2/ci_matrix", # Temporary, don't merge this
```
VictorKoenders: I'll remove this once the PR is approved (This time for sure I swear)

ghost (approved):

.github/workflows/rust.yml
VictorKoenders: I totally did not almost forget this

PR #422, opened 2021-10-25
functions to enable encoding/decoding serde types

VictorKoenders:
This is an initial attempt at supporting serde in bincode 2.

I did have to make separate functions for all serde-related functionality. This is possibly preventable.

Implementing `Serialize` was relatively easy.

For `Deserialize` and `DeserializeOwned` I ended up making 2 functions, which support `BorrowDecoder` and `Decoder` respectfully. These also have 2 separate functions right now, we might be able to merge this into one function.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ description = "A binary serialization / deserialization strategy for transformin
 edition = "2021"
 
 [features]
-default = ["std", "derive", "atomic", "serde"] # If you see this in a PR, I've forgotten to remove it
+default = ["std", "derive", "atomic"]
 std = ["alloc"]
 alloc = []
 atomic = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,15 +23,20 @@ description = "A binary serialization / deserialization strategy for transformin
 edition = "2021"
 
 [features]
-default = ["std", "derive", "atomic"]
+default = ["std", "derive", "atomic", "serde"] # If you see this in a PR, I've forgotten to remove it
 std = ["alloc"]
 alloc = []
 atomic = []
 derive = ["bincode_derive"]
 
+# BlockedTODO: https://github.com/rust-lang/cargo/issues/8832
+# We really want these features to automatically be enabled when both "serde" and either "alloc"/"std" is enabled
+# But this is currently not possible
+serde = ["std", "serde_incl", "serde_incl/std"]
+
 [dependencies]
 bincode_derive = { path = "derive", version = "2.0.0-alpha.0", optional = true }
-serde = { version = "1.0.130", optional = true }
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -23,15 +23,20 @@ description = "A binary serialization / deserialization strategy for transformin
 edition = "2021"
 
 [features]
-default = ["std", "derive", "atomic"]
+default = ["std", "derive", "atomic", "serde"] # If you see this in a PR, I've forgotten to remove it
```
ghost: I have seen this in a PR

src/features/serde/de_borrowed.rs
```diff
@@ -0,0 +1,401 @@
...
+        V: serde_incl::de::Visitor<'de>,
+    {
+        visitor.visit_byte_buf(Decode::decode(&mut self.de)?)
+    }
+
+    fn deserialize_option<V>(mut self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: serde_incl::de::Visitor<'de>,
+    {
+        let val = Option::<()>::decode(&mut self.de)?;
```
ghost: Does this really only get called for `Option<()>`?

ghost (commented):
I feel like this should have an `impl Encode for T where T: Serialize` and same for `Decode`. With the current implementation you can't bring a struct that only implements `Serialize` into a struct you want to implement `Encode` on

src/features/serde/de_borrowed.rs
VictorKoenders: I'm using `Option<()>` because I only care about the `Some`/`None` variant, and I know `()` won't actually read from the stream.
Alternatively I could ready `u8` here, but then we have multiple places where we're decoding an `Option`.

VictorKoenders:
I've opened a separate issue for having bincode `Decode/Encode` work together with serde's traits here: #433

tests/serde.rs
```diff
@@ -29,3 +33,95 @@ fn test_serde_round_trip() {
...
+    pub b: u32,
+    pub str: &'a str,
+}
+
+#[test]
+fn test_serialize_deserialize_borrowed_data() {
+    let input = SerdeWithBorrowedData {
+        a: 5,
+        b: 5,
+        str: "Hello world",
```
ghost, line 51: this... works? I thought `str` was a keyword. Oh well /shrug

ghost (approved):

tests/serde.rs
VictorKoenders: [Don't worry about it](https://github.com/rust-lang/rust/blob/38fa66aac3cb1790bde8f76014db6ccd62f70244/src/test/ui/weird-exprs.rs#L87-L107)

Issue #427, opened 2021-10-26
bincode::Encode seems to fail pretty hard and tries to modify the doc comment


haata:
bincode-2.0.0-alpha.0

```rust
/// HID-IO Packet Buffer Struct
///
/// # Remarks
/// Used to store HID-IO data chunks. Will be chunked into individual packets on transmission.
#[repr(C)]
#[derive(PartialEq, Clone, Debug, bincode::Encode)]
#[cfg_attr(feature = "defmt-impl", derive(defmt::Format))]
pub struct HidIoPacketBuffer<const H: usize> {
    /// Type of packet (Continued is automatically set if needed)
    pub ptype: HidIoPacketType,
    /// Packet Id
    pub id: HidIoCommandId,
    /// Packet length for serialization (in bytes)
    pub max_len: u32,
    /// Payload data, chunking is done automatically by serializer
    pub data: Vec<u8, H>,
    /// Set False if buffer is not complete, True if it is
    pub done: bool,
}
```

```
error: Invalid rust syntax
   --> hid-io-protocol/src/lib.rs:158:1
    |
158 | ///
    | ^^^
```

Some notes:
- Embedded target, no alloc, no_std
- All features disabled, except for derive
- Vec is actually a heapless::Vec  
- Same problem if I try it on a simple enum
- defmt::Format is a serialized logging format for embedded targets used for debugging

```rust
#[repr(u32)]
#[derive(PartialEq, Clone, Copy, Debug, IntoPrimitive, TryFromPrimitive, bincode::Encode)]
#[cfg_attr(feature = "defmt-impl", derive(defmt::Format))]
/// Requests for to perform a specific action
pub enum HidIoCommandId {
    SupportedIds = 0x00,
    GetInfo = 0x01,
    TestPacket = 0x02,
    ResetHidIo = 0x03,
    Reserved = 0x04, // ... 0x0F

    GetProperties = 0x10,
    KeyState = 0x11,
    KeyboardLayout = 0x12,
    KeyLayout = 0x13,
    KeyShapes = 0x14,
    LedLayout = 0x15,
    FlashMode = 0x16,
    UnicodeText = 0x17,
    UnicodeState = 0x18,
    HostMacro = 0x19,
    SleepMode = 0x1A,

    KllState = 0x20,
    PixelSetting = 0x21,
    PixelSet1c8b = 0x22,
    PixelSet3c8b = 0x23,
    PixelSet1c16b = 0x24,
[...]

PR #428, opened 2021-10-26
Fixes for 427

VictorKoenders:
#427 uncovered some issues.

1. if there is a newline in a doc comment, rust's token stream feeds us 2x `'#'` in a row with no attribute group. `bincode_derive` now handles this correctly
2. `bincode_derive` now supports min_const_generics
3. `bincode_derive` now supports c-style enums where the variants have a fixed value

For 3 I have made it so that bincode_derive outputs code like this:
<details>
<summary>CStyleEnum output</summary>

```rust
enum CStyleEnum {
    A = 1,
    B = 2,
    C,
    D = 5,
    E,
}


impl bincode::de::Decode for CStyleEnum {
    fn decode<D: bincode::de::Decoder>(
        mut decoder: D,
    ) -> Result<Self, bincode::error::DecodeError> {
        let variant_index = <u32 as bincode::de::Decode>::decode(&mut decoder)?;
        match variant_index {
            1 => Ok(Self::A {}),
            2 => Ok(Self::B {}),
            x if x == 2 + 1u32 => Ok(Self::C {}),
            5 => Ok(Self::D {}),
            x if x == 5 + 1u32 => Ok(Self::E {}),
            variant => Err(bincode::error::DecodeError::UnexpectedVariant {
                found: variant,
                type_name: "CStyleEnum",
                allowed: bincode::error::AllowedEnumVariants::Allowed(&[
                    1,
                    2,
                    2 + 1u32,
                    5,
                    5 + 1u32,
                ]),
            }),
        }
    }
}

impl bincode::enc::Encode for CStyleEnum {
    fn encode<E: bincode::enc::Encoder>(
        &self,
        mut encoder: E,
    ) -> core::result::Result<(), bincode::error::EncodeError> {
        match self {
            Self::A => {
                <u32 as bincode::enc::Encode>::encode(&(1), &mut encoder)?;
            }
            Self::B => {
                <u32 as bincode::enc::Encode>::encode(&(2), &mut encoder)?;
            }
            Self::C => {
                <u32 as bincode::enc::Encode>::encode(&(2 + 1u32), &mut encoder)?;
            }
            Self::D => {
                <u32 as bincode::enc::Encode>::encode(&(5), &mut encoder)?;
            }
            Self::E => {
                <u32 as bincode::enc::Encode>::encode(&(5 + 1u32), &mut encoder)?;
            }
        }
        Ok(())
    }
}
```

</details>

### DIFF
--- a/derive/src/derive_enum.rs
+++ b/derive/src/derive_enum.rs
@@ -1,4 +1,4 @@
-use crate::generate::{FnSelfArg, Generator};
+use crate::generate::{FnSelfArg, Generator, StreamBuilder};
 use crate::parse::{EnumVariant, Fields};
 use crate::prelude::*;
 use crate::Result;
@@ -10,9 +10,15 @@ pub struct DeriveEnum {
 }
 
 impl DeriveEnum {
-    pub fn generate_encode(self, generator: &mut Generator) -> Result<()> {
-        let DeriveEnum { variants } = self;
+    fn iter_fields(&self) -> EnumVariantIterator {
+        EnumVariantIterator {
+            idx: 0,
+            last_val: None,
+            variants: &self.variants,
+        }
+    }
 
+    pub fn generate_encode(self, generator: &mut Generator) -> Result<()> {
         generator
             .impl_for("bincode::enc::Encode")
             .unwrap()
@@ -25,7 +31,7 @@ impl DeriveEnum {
                 fn_body.ident_str("match");
                 fn_body.ident_str("self");
                 fn_body.group(Delimiter::Brace, |match_body| {
-                    for (variant_index, variant) in variants.into_iter().enumerate() {
+                    for (variant_index, variant) in self.iter_fields() {
                         // Self::Variant
                         match_body.ident_str("Self");
                         match_body.puncts("::");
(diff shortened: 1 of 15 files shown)


VictorKoenders:
@haata the only thing that is left of #427 is that `heapless::Vec` does not support `Decode/Encode` for now. I think it's be best to use a wrapper for this that implements Decode/Encode. Something like:

```rust

#[repr(transparent)]
struct BincodeVec<T, const N: usize>(heapless::Vec<T, N>);

impl<T: Decode + Default, const N: usize> Decode for BincodeVec<T, N> {
    fn decode<D: bincode::de::Decoder>(
        mut decoder: D,
    ) -> Result<Self, bincode::error::DecodeError> {
        let len = usize::decode(&mut decoder)?;
        let mut vec = heapless::Vec::default();
        for _ in 0..len {
            // TODO: handle the heapless error?
            let _ = vec.push(T::decode(&mut decoder)?);
        }
        Ok(Self(vec))
    }
}

impl<T: Encode, const N: usize> Encode for BincodeVec<T, N> {
    fn encode<E: bincode::enc::Encoder>(
        &self,
        mut encoder: E,
    ) -> Result<(), bincode::error::EncodeError> {
        self.0.len().encode(&mut encoder)?;
        for val in self.0.iter() {
            val.encode(&mut encoder)?;
        }
        Ok(())
    }
}
```

derive/src/derive_enum.rs
```diff
@@ -153,45 +230,79 @@ impl DeriveEnum {
...
+        self.idx += 1;
+
+        let tokens = if let Fields::Integer(lit) = &variant.fields {
+            let tree = TokenTree::Literal(lit.clone());
+            self.last_val = Some((lit.clone(), 0));
+            vec![tree]
+        } else if let Some((lit, add)) = self.last_val.as_mut() {
+            *add += 1;
+            vec![
+                TokenTree::Literal(lit.clone()),
```
ghost, line 298: instead of doing this could we parse the literal into a number and then do the addition ourselves, spitting out the resulting number?

VictorKoenders: Unfortunately not, https://doc.rust-lang.org/stable/proc_macro/struct.Literal.html does not have any way to get the actual value, unless we want to do a `lit.to_string().parse().unwrap()` or something.

ghost (approved):

PR #430, opened 2021-11-05
Generate qualified Result type in derive

andrenth:
This prevents problems with generated code when there's an alias for a crate-specific `Result` type, i.e. `type Result<T> = std::result::Result<T, MyError>`.


VictorKoenders (approved):

VictorKoenders:
Thanks!

Just out of sheer interest, how easy did you find it to go through the `bincode_derive` crate and figure out where you had to change things?

andrenth:
It was pretty easy to find, but honestly I didn't even have to go through the code, because a `grep Result derive/src/*` pretty much told me what to do :)

PR #432, opened 2021-11-07
split off BorrowDecode from Decode in bincode_derive

VictorKoenders:
We've discussed splitting off `Decode` from `BorrowDecode` derive macro before. This PR actualizes that.

This should offer people a solution to issues where the type implements `BorrowDecode` while people are explicitly requesting `Decode`. It does not fix the issue around `Cow<T>` always needing `Decode`, we can't fix that until specialization implementation.

### DIFF
--- a/derive/src/derive_enum.rs
+++ b/derive/src/derive_enum.rs
@@ -168,7 +168,7 @@ impl DeriveEnum {
         let enum_name = generator.target_name().to_string();
 
         generator
-            .impl_for("bincode::de::Decode")
+            .impl_for("bincode::Decode")
             .unwrap()
             .generate_fn("decode")
             .with_generic("D", ["bincode::de::Decoder"])
@@ -177,7 +177,7 @@ impl DeriveEnum {
             .body(|fn_builder| {
                 fn_builder
                     .push_parsed(
-                        "let variant_index = <u32 as bincode::de::Decode>::decode(&mut decoder)?;",
+                        "let variant_index = <u32 as bincode::Decode>::decode(&mut decoder)?;",
                     )
                     .unwrap();
                 fn_builder.push_parsed("match variant_index").unwrap();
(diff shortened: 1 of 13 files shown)


VictorKoenders:
I've taken the liberty to also re-export the `Decode` and `Encode` trait in the root of the crate, as these are used often

derive/src/derive_enum.rs
```diff
@@ -86,112 +86,116 @@ impl DeriveEnum {
...
-                                let is_tuple = matches!(variant.fields, Fields::Tuple(_));
-                                for (idx, field) in variant.fields.names().into_iter().enumerate() {
-                                    if is_tuple {
-                                        variant_body.lit_usize(idx);
-                                    } else {
-                                        variant_body.ident(field.unwrap_ident().clone());
-                                    }
-                                    variant_body.punct(':');
-                                    variant_body.push_parsed("bincode::de::BorrowDecode::borrow_decode(&mut decoder)?,").unwrap();
+        // enum has no lifetimes, implement Decode
```
ghost: This comment is no longer needed

derive/src/derive_struct.rs
```diff
@@ -36,74 +36,76 @@ impl DeriveStruct {
     }
 
     pub fn generate_decode(self, generator: &mut Generator) -> Result<()> {
+        // Remember to keep this mostly in sync with generate_borrow_decode
         let DeriveStruct { fields } = self;
 
-        if generator.has_lifetimes() {
-            // struct has a lifetime, implement BorrowDecode
+        // struct has no lifetimes, implement Decode
```
ghost: same here

derive/src/derive_enum.rs
VictorKoenders: Removed

derive/src/derive_struct.rs
VictorKoenders: Removed

ghost (approved):

PR #434, opened 2021-11-08
Allow serde types to be Decode/Encoded

### LINKED ISSUE #433: Figure out bincode Decode/Encode interop with serde
Currently bincode 2 only allows either `Decode/Encode`, or serde's traits. It would be nice to have a way to use serde types while using `Decode/Encode`

Some ideas:
- Add an `#[serde_interop]` attribute to `bincode_derive`
- Add a `WithSerde<T>` that implements `Decode/Encode`
- Disable the default impl for `Decode/Encode` and replace it with something like:
  - `impl Decode for T where T: Deserialize`
  - `impl Encode for T where T: Serialize`

VictorKoenders:
Adds an attribute that users can add to struct fields
```rust
#[derive(bincode::Encode, bincode::Decode)]
struct Foo {
    #[bincode(with_serde)]
    serde: SerdeType
}
```

Also adds a struct that will wrap any serde type and implements `Decode` and `Encode` on it

Closes #433

### DIFF
--- a/derive/src/derive_struct.rs
+++ b/derive/src/derive_struct.rs
@@ -1,5 +1,5 @@
 use crate::generate::Generator;
-use crate::parse::Fields;
+use crate::parse::{FieldAttribute, Fields};
 use crate::prelude::Delimiter;
 use crate::Result;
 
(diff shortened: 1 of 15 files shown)


derive/src/derive_struct.rs
```diff
@@ -21,12 +21,21 @@ impl DeriveStruct {
...
-                    fn_body
-                        .push_parsed(format!(
-                            "bincode::enc::Encode::encode(&self.{}, &mut encoder)?;",
-                            field.to_string()
-                        ))
-                        .unwrap();
+                    if field.has_field_attribute(FieldAttribute::WithSerde) {
+                        fn_body
+                            .push_parsed(format!(
+                                "bincode::__serde_encode_field(&self.{}, &mut encoder)?;",
```
ghost: Instead of calling a function that is doc hidden, couldn't we just wrap the field in the public `SerdeToBincode` and then call encode/decode on that?

VictorKoenders: This would work for `Encode` and `Decode`, but not for `BorrowDecode`.

I think I'll make a wrapper `bincode::serde::Compat` (Decode/Encode) and `bincode::serde::BorrowCompat` (BorrowDecode/Encode)

VictorKoenders: Done!

ghost (approved):

PR #439, opened 2021-11-29
Feature/config limit

VictorKoenders:
Adds a limit to configuration and keeps track of how many bytes are going to be read, returning a `DecodeError::LimitExceeded` if this exceeds the amount of bytes read.

The reason this is set up like this is because we use a lot of `Container::with_capacity` when decoding containers. The limit exists primarily to not allocate too much value, for example when data is garbage, we don't want to accidentally allocate 10GB of memory when the data is invalid.

I've added an `unclaim` system as well. As per the `Decoder` comments:

> When decoding container types, a typical implementation would claim to read `len * size_of::<T>()` bytes.
> This is to ensure that bincode won't allocate several GB of memory while constructing the container.
> 
> Because the implementation claims `len * size_of::<T>()`, but then has to decode each `T`, this would be marked
> as double. This function allows us to un-claim each `T` that gets decoded.
> 
> We cannot check if `len * size_of::<T>()` is valid without claiming it, because this would mean that if you have
> a nested container (e.g. `Vec<Vec<T>>`), it does not know how much memory is already claimed, and could easily
> allocate much more than the user intends.

### DIFF
--- a/src/de/decoder.rs
+++ b/src/de/decoder.rs
@@ -2,7 +2,7 @@ use super::{
     read::{BorrowReader, Reader},
     BorrowDecoder, Decoder,
 };
-use crate::{config::Config, utils::Sealed};
+use crate::{config::Config, error::DecodeError, utils::Sealed};
 
 /// A Decoder that reads bytes from a given reader `R`.
 ///
@@ -24,12 +24,17 @@ use crate::{config::Config, utils::Sealed};
 pub struct DecoderImpl<R, C: Config> {
     reader: R,
     config: C,
+    bytes_read: usize,
 }
 
 impl<R: Reader, C: Config> DecoderImpl<R, C> {
     /// Construct a new Decoder
     pub fn new(reader: R, config: C) -> DecoderImpl<R, C> {
-        DecoderImpl { reader, config }
+        DecoderImpl {
+            reader,
+            config,
+            bytes_read: 0,
+        }
     }
 }
 
@@ -54,5 +59,31 @@ impl<R: Reader, C: Config> Decoder for DecoderImpl<R, C> {
 
     fn config(&self) -> &Self::C {
         &self.config
+    }
+
+    fn claim_bytes_read(&mut self, n: usize) -> Result<(), DecodeError> {
+        // C::LIMIT is a const so this check should get compiled away
+        if let Some(limit) = C::LIMIT {
+            // Make sure we don't accidentally overflow `bytes_read`
+            self.bytes_read = self
+                .bytes_read
+                .checked_add(n)
+                .ok_or(DecodeError::LimitExceeded)?;
+            if self.bytes_read > limit {
+                Err(DecodeError::LimitExceeded)
+            } else {
+                Ok(())
+            }
+        } else {
+            Ok(())
+        }
+    }
+
+    fn unclaim_bytes_read(&mut self, n: usize) {
+        // C::LIMIT is a const so this check should get compiled away
+        if C::LIMIT.is_some() {
+            // We should always be claiming more than we unclaim, so this should never underflow
+            self.bytes_read -= n;
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


VictorKoenders:
TODO(self):
- [x] Add tests to see if the limit actually gets hit
- [ ] Maybe add a check `SliceReader` to see if the slice is larger than the current limit, and error immediately?

src/de/decoder.rs
```diff
@@ -55,4 +60,30 @@ impl<R: Reader, C: Config> Decoder for DecoderImpl<R, C> {
     fn config(&self) -> &Self::C {
         &self.config
     }
+
+    fn claim_bytes_read(&mut self, n: usize) -> Result<(), DecodeError> {
```
ghost, line 65: inline these

src/de/impls.rs
```diff
@@ -405,6 +424,9 @@ where
             }
         }
 
+        decoder.claim_bytes_read(N * core::mem::size_of::<T>())?;
```
ghost: I think `size_of` should just be able to be written as `core::mem::size_of::<[T; N]>()` unless const generics die there

src/de/impls.rs
```diff
@@ -415,8 +437,11 @@ where
             let res = unsafe { ptr.read() };
             Ok(res)
         } else {
-            let result =
-                super::impl_core::collect_into_array(&mut (0..N).map(|_| T::decode(&mut decoder)));
+            let result = super::impl_core::collect_into_array(&mut (0..N).map(|_| {
+                // See the documentation on `unclaim_bytes_read` as to why we're doing this here
+                decoder.unclaim_bytes_read(core::mem::size_of::<T>());
```
ghost, line 466: rather than unclaiming in a loop, we could just unclaim right after claiming. That should reduce loop maths.

src/features/impl_alloc.rs
```diff
@@ -52,8 +52,13 @@ where
 {
     fn decode<D: Decoder>(mut decoder: D) -> Result<Self, DecodeError> {
         let len = crate::de::decode_slice_len(&mut decoder)?;
+        decoder.claim_container_read::<T>(len)?;
```
ghost, line 55: why is this used sometimes and other times (like the array impl) done manually? Also, same comment here about immediately unclaiming

ghost (changes requested):

src/de/impls.rs
VictorKoenders: The reason I unclaim per item is because of nested containers.

Consider the situation where there is data type `Vec<Vec<u8>>`, where each `Vec` is 100 entries. This means the entire container would allocate 10kb of memory.

If our limit is set to 2500, the current impl would:
- Allocate the outer `Vec<Vec<_>>` (100 * 24), count = 2400
- Deallocate 1 inner `Vec<u8>` (24 bytes), count = 2376
- Allocate the inner `Vec<u8>` (100 bytes), count = 2476
- Deallocate 1 inner `Vec<u8>` (24 bytes), count = 2452
- Allocate the inner `Vec<u8>` (100 bytes), count = 2552

This would fail at the 3rd allocation

Whereas if we unclaim the entire vec immediately, we would get:
- Allocate the outer `Vec<Vec<_>>` (100 * 24), count = 2400
- Deallocate this, count = 0
- 25x:
  - Allocate the inner `Vec<u8>` (100 bytes), count = 100, 200, etc
 
This would fail at the 25th allocation

src/features/impl_alloc.rs
VictorKoenders: This is specifically for containers (e.g. `Vec`, `HashMap`, etc). An array `[T; N]` is in my eyes not a container

src/de/decoder.rs
VictorKoenders: Added a benchmark and it seems to increase speed between 0 and 2%, so I guess that's something

src/de/impls.rs
VictorKoenders: Done

ghost (approved):

PR #443, opened 2021-12-04
Extract virtue

VictorKoenders:
This PR extracts the derive framework from `bincode-derive` into its own crate. This crate is https://docs.rs/virtue.

Additionally fixes some minor issues
- `src/impl_core.rs` now has some clippy warnings. Because this code is copied from `core` I've just added a `#![allow(clippy::needless_borrow)]` at the top
- Clippy has a new warning to replace `core::mem::replace(&mut self.slice, &mut [])` with `core::mem::take(&mut self.slice)`, so I've implemented that
- `#![crate_type = "dylib"]` is now deprecated, so I've removed it


src/de/impl_core.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(unused_unsafe)]
+#![allow(clippy::needless_borrow)]
```
ghost: is this still needed after your clippy fix?

VictorKoenders: Nope, removed!

ghost (approved):

PR #445, opened 2021-12-09
[Breaking change] Made all `decode_from_slice` also return the number of bytes read

VictorKoenders:
Currently there is no way to know how many bytes are read when decoding from a slice. This will make it harder to read multiple entries from a blob without encoding the length. This PR makes all `decode_from_slice` functions return the amount of bytes read. 

This is a breaking change, but we're still on alpha so this should be okay.

Also fixes some of the new clippy warnings. This will conflict with #443 but we'll fix that when we get there

### DIFF
--- a/benches/varint.rs
+++ b/benches/varint.rs
@@ -14,7 +14,7 @@ fn slice_varint_u8(c: &mut Criterion) {
 
     c.bench_function("slice_varint_u8", |b| {
         b.iter(|| {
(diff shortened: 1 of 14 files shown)


ghost (approved):
I don't really like returning the tuple, but the only alternative would be to make the `decode_from_slice` function take a mutable reference, which is more restrictive than it need be.

PR #446, opened 2021-12-10
Impl BorrowDecode for Option<&[u8]> and Option<&str>

songzhi:
Since those two types are encodable, implements `BorrowDecode` for them is nature.


VictorKoenders:
This makes a lot of sense, we're only implementing `Decode` for `Option<T> where T: Decode`. I tried implementing it for `Option<T> where T: BorrowDecode` but that breaks a lot of things.

Clippy has some new warnings, they will be fixed in #447. After that is merged can you rebase? PR looks good to me

VictorKoenders (approved):

PR #447, opened 2021-12-10
Fixed clippy warnings

VictorKoenders:
Clippy has some new warnings, that were being fixed in multiple PRs. Fixing it here so the rest isn't blocked on this.


ghost (approved):
Looks good. We should probably considering pinning the clippy version so this doesn’t bite us in the future.

PR #457, opened 2021-12-22
Fix failed varint bench

ygf11:


VictorKoenders:
Oops I guess we missed this in #445, thanks!

VictorKoenders (approved):

PR #462, opened 2022-01-08
Fixed derive impl on an empty enum

VictorKoenders:
bincode-derive failed on an empty enum. This PR fixes this.

Apparently if you implement a trait on an enum, an empty match statement is not sufficient. Rust seems to think the enum is populated somehow. Adding a `unreachable!()` fixed this

```rs
impl bincode::enc::Encode for BorrowedEmptyEnum {
    fn encode<E: bincode::enc::Encoder>(
        &self,
        mut encoder: E,
    ) -> core::result::Result<(), bincode::error::EncodeError> {
        match self {
            _ => core::unreachable!(), // new
        }
    }
}
```

Update: Made the `Decode` not read the byte and return a specific error for this situation

```rs
impl bincode::Decode for EmptyEnum {
    fn decode<D: bincode::de::Decoder>(
        mut decoder: D,
    ) -> core::result::Result<Self, bincode::error::DecodeError> {
        core::result::Result::Err(bincode::error::DecodeError::EmptyEnum {
            type_name: core::any::type_name::<Self>(),
        })
    }
}
```

<details>

<summary>old</summary>

Additionally the `self.variants.len() - 1` would underflow if there are no variants. `saturating_sub(1)` sets it correctly to 0

```rs
impl bincode::Decode for EmptyEnum {
    fn decode<D: bincode::de::Decoder>(
        mut decoder: D,
    ) -> core::result::Result<Self, bincode::error::DecodeError> {
        let variant_index = <u32 as bincode::Decode>::decode(&mut decoder)?;
        match variant_index {
            variant => Err(bincode::error::DecodeError::UnexpectedVariant {
                found: variant,
                type_name: "EmptyEnum",
                allowed: bincode::error::AllowedEnumVariants::Range { min: 0, max: 0 },
            }),
        }
    }
}
```

</details>

### DIFF
--- a/derive/src/derive_enum.rs
+++ b/derive/src/derive_enum.rs
@@ -35,6 +35,9 @@ impl DeriveEnum {
                 fn_body.ident_str("match");
                 fn_body.ident_str("self");
                 fn_body.group(Delimiter::Brace, |match_body| {
+                    if self.variants.is_empty() {
+                        self.add_empty_variant_case(match_body)?;
+                    }
                     for (variant_index, variant) in self.iter_fields() {
                         // Self::Variant
                         match_body.ident_str("Self");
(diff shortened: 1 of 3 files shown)


derive/src/derive_enum.rs
```diff
@@ -168,7 +177,7 @@ impl DeriveEnum {
                     // no fixed values, implement a range
                     variant_inner.push_parsed(format!(
                         "bincode::error::AllowedEnumVariants::Range {{ min: 0, max: {} }}",
-                        self.variants.len() - 1
+                        self.variants.len().saturating_sub(1)
```
ghost: Wouldn't this make more sense as an error dedicated specifically to "Tried to deserialize an uninhabited enum?"

VictorKoenders: I agree, I also made is to that it doesn't read data from the reader, because this isn't needed

```rs
impl bincode::Decode for EmptyEnum {
    fn decode<D: bincode::de::Decoder>(
        mut decoder: D,
    ) -> core::result::Result<Self, bincode::error::DecodeError> {
        core::result::Result::Err(bincode::error::DecodeError::EmptyEnum {
            type_name: core::any::type_name::<Self>(),
        })
    }
}
```

Added a test for this as well

ghost (approved):

PR #465, opened 2022-01-08
Fix overflow error when deserializing invalid Duration

ghost:


tests/basic_types.rs
```diff
@@ -219,3 +219,44 @@ fn test_array() {
...
+fn test_duration_out_of_range() {
+    let mut input = [0u8; 14];
+
+    bincode::encode_into_slice(&(u64::MAX, u32::MAX), &mut input, Configuration::standard())
+        .unwrap();
+
+    let result: Result<(std::time::Duration, usize), _> =
+        bincode::decode_from_slice(&mut input, Configuration::standard());
+
+    match result {
```
VictorKoenders: You could use `.unwrap_err()` here

VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #468, opened 2022-01-08
Add fuzzing harness, try to decode into various types

ghost:
This isn't everything that can be tested, but it's a good starting point, and already found two issues.


VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #471, opened 2022-01-12
Switch Encode to take &mut E

### LINKED ISSUE #459: ICE: `thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0'`
See: https://github.com/kdy1/wasm-perf/pull/1

I filed [an issue on rust issue tracker](https://github.com/rust-lang/rust/issues/92470), but I think it can be workarounded for now.

Stack:

```

query stack during panic:
#0 [codegen_fulfill_obligation] checking if `bincode::config::internal::InternalIntEncodingConfig` fulfil
ls its obligations
#1 [resolve_instance] resolving instance `<_ as bincode::config::internal::InternalIntEncodingConfig>::IN
T_ENCODING`
```

ghost:
Fixes #459

### DIFF
--- a/derive/src/derive_enum.rs
+++ b/derive/src/derive_enum.rs
@@ -29,7 +29,7 @@ impl DeriveEnum {
             .generate_fn("encode")
             .with_generic("E", ["bincode::enc::Encoder"])
             .with_self_arg(FnSelfArg::RefSelf)
-            .with_arg("mut encoder", "&mut E")
+            .with_arg("encoder", "&mut E")
             .with_return_type("core::result::Result<(), bincode::error::EncodeError>")
             .body(|fn_body| {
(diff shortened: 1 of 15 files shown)


derive/src/derive_enum.rs
```diff
@@ -29,7 +29,7 @@ impl DeriveEnum {
             .generate_fn("encode")
             .with_generic("E", ["bincode::enc::Encoder"])
             .with_self_arg(FnSelfArg::RefSelf)
-            .with_arg("mut encoder", "E")
+            .with_arg("mut encoder", "&mut E")
```
VictorKoenders: This can be `"encoder"`

VictorKoenders (approved):

PR #481, opened 2022-01-19
Moved Configuration::standard() and ::legacy() to the config module

VictorKoenders:


src/config.rs
```diff
@@ -47,42 +46,40 @@ pub struct Configuration<E = LittleEndian, I = Varint, A = SkipFixedArrayLength,
...
+        _e: PhantomData,
+        _i: PhantomData,
+        _a: PhantomData,
+        _l: PhantomData,
     }
+}
 
+impl<E, I, A, L> Configuration<E, I, A, L> {
     /// Makes bincode encode all integer types in big endian.
     pub const fn with_big_endian(self) -> Configuration<BigEndian, I, A> {
```
ghost: Unrelated to this PR in particular, but I just noticed that this doesn't specify `L` in the return type. Does this mean we are erasing the limits?

VictorKoenders: Yes...

fixed and added a comment on adding more options

ghost (approved):

PR #491, opened 2022-01-25
Return an error if a decoded slice length doesn't fit into usize

poljar:
Bincode encodes a slice length, which is an usize, as an u64. When such an encoded slice length needs to be decoded it again uses an u64 but critically it truncates it into an usize.

An usize is architecture dependent, it is the size how many bytes it takes to reference any location in memory. The most common sizes for an usize are 64, 32, and 16 bit.

This might lead to silent data loss if the architecture that encoded the slice differs from the architecture that decoded the slice, i.e. if we go from a 64 bit architecture to a 32 or 16 bit one.

Since bincode aims to be suitable for storage, aiming to support the exchange of data between different architectures silently truncating such slice lenghts should be avoided.

This patch changes the behaviour to error out if we try to decode an slice lenght that can't fit into the current usize type.

### DIFF
--- a/src/de/impls.rs
+++ b/src/de/impls.rs
@@ -180,7 +180,7 @@ impl Decode for usize {
 
                 value
                     .try_into()
-                    .map_err(|_| DecodeError::SliceLength(value))
+                    .map_err(|_| DecodeError::OutsideUsizeRange(value))
             }
         }
     }
--- a/src/de/impls.rs
+++ b/src/de/impls.rs
@@ -172,10 +172,15 @@ impl Decode for usize {
             IntEncoding::Fixed => {
                 let mut bytes = [0u8; 8];
                 decoder.reader().read(&mut bytes)?;
-                Ok(match D::C::ENDIAN {
+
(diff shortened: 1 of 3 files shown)


ghost (approved):

VictorKoenders:
Quick scan through the source code to see if we have some more cases where we cast to usize, found some:
- https://github.com/bincode-org/bincode/blob/8c3dd01b89fe9edd2ce0c31f201b4f372f5a9663/src/de/impls.rs#L178
- Comment here: https://github.com/bincode-org/bincode/blob/8c3dd01b89fe9edd2ce0c31f201b4f372f5a9663/src/de/mod.rs#L104
- array indices when decoding chars, which is probably fine.
- tuples, which is also fine up until `u16::MAX` entries

Can you add this check to the 2 places above?

poljar:
> Can you add this check to the 2 places above?

Did so, I also renamed the error variant since it isn't length specific anymore.

VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #493, opened 2022-01-28
Support slice lengths to be encoded as an u32

poljar:
Slice lengths (an usize) are encoded in bincode as a u64 value. Given that the width of usize is platform-specific, an usize value encoded on a platform with a wider usize has the potential for decoding to fail on a platform with a narrower usize. If we know the usize width of the target decoding platform at encoding time, it could be preferable to fail at encoding time instead, giving us the possibility to detect this error.

This commit introduces the ability to choose, at encoding time, between slices being 32 or 64 bits wide. These widths are chosen for being the two most common slice length type widths on common architectures. If a slice length value is given that is out of range for the chosen slice length type, an error will happen at encoding time.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -30,6 +30,7 @@ use core::marker::PhantomData;
 /// - [with_little_endian] and [with_big_endian]
 /// - [with_fixed_int_encoding] and [with_variable_int_encoding]
 /// - [skip_fixed_array_length] and [write_fixed_array_length]
+/// - [with_u32_slice_length_encoding] and [with_u64_slice_length_encoding]
 ///
 ///
 /// [with_little_endian]: #method.with_little_endian
@@ -38,12 +39,21 @@ use core::marker::PhantomData;
 /// [with_variable_int_encoding]: #method.with_variable_int_encoding
 /// [skip_fixed_array_length]: #method.skip_fixed_array_length
 /// [write_fixed_array_length]: #method.write_fixed_array_length
+/// [with_u32_slice_length_encoding]: #method.with_u32_slice_length_encoding
+/// [with_u64_slice_length_encoding]: #method.with_u64_slice_length_encoding
 #[derive(Copy, Clone)]
-pub struct Configuration<E = LittleEndian, I = Varint, A = SkipFixedArrayLength, L = NoLimit> {
+pub struct Configuration<
+    E = LittleEndian,
+    I = Varint,
+    A = SkipFixedArrayLength,
+    L = NoLimit,
+    LE = U64SliceLengthEncoding,
+> {
     _e: PhantomData<E>,
     _i: PhantomData<I>,
     _a: PhantomData<A>,
     _l: PhantomData<L>,
+    _le: PhantomData<LE>,
 }
 
 /// The default config for bincode 2.0. By default this will be:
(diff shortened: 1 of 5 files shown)


VictorKoenders:
I'm not sure if this is actually needed. The smallest slice we can create is a `&[u8]`. If we have `u32::MAX` entries in this slice, this will be 4 gb of memory. 32 bit platforms by definition cannot handle more than 4gb of memory. So if you send `u32::MAX + 1` entries, the 32 bit platform is unable to process it anyway.

32 bit platforms can handle more data by processing a stream, however bincode is not a protocol that supports streaming without a custom implementation anyway.

The only thing I can imagine is that on 32 bit platforms, it might be slightly faster to decode a `u32` than an `u64`. If this is happening we might want to decode `usize` as `u32` all the time anyway, and not introduce a config option.

VictorKoenders:
On the encoders side you can catch this issue easily by adding a limit of 4gb
```rs
const 4GB: usize = u32::MAX as usize;
bincode::config::standard().with_limit::<4GB>()
```

poljar:
> I'm not sure if this is actually needed. The smallest slice we can create is a `&[u8]`.

That's not quite correct, a bool could use less than an `u8`, though currently an i8 seems to be used by LLVM[[1]]:

> Rust emits i1 to LLVM for bool and relies on whatever it produces. LLVM uses i8 (one byte) to represent i1 in memory for all the platforms supported by Rust for now.

For completeness sake, we also have zero sized types in Rust, though I don't think that they would be useful inside a slice.

> If we have `u32::MAX` entries in this slice, this will be 4 gb of memory. 32 bit platforms by definition cannot handle more than 4gb of memory. So if you send `u32::MAX + 1` entries, the 32 bit platform is unable to process it anyway.

That's also not quite correct and more important than the claim above. Thanks to [PAE] a 32 bit Kernel can access more than 4GB. A single process is still limited to the 32 bit virtual address space, but various tricks, i.e. using mmap, could be utilized to access more than 4GB in a single process as well.

> 32 bit platforms can handle more data by processing a stream, however bincode is not a protocol that supports streaming without a custom implementation anyway.

As mentioned, one could use mmap or shmfs/tmpfs to access more than 4GB on a 32 bit system, under the assumption that [PAE] is utilized.

> The only thing I can imagine is that on 32 bit platforms, it might be slightly faster to decode a `u32` than an `u64`. If this is happening we might want to decode `usize` as `u32` all the time anyway, and not introduce a config option.

I have not measured this but I assume that it would be faster regardless considering that there's 4 bytes less to decode. That being said, I don't think it makes sense to limit 64 bit systems in such a way.

As for the reason why I specifically want this, the cryptographic library used in Matrix land, [libolm], uses a bincode compatible format but uses u32 for slice lengths. We're currently doing a Rust rewrite of libolm called [vodozemac], it would be neat if I could use bincode to read the data that libolm produces.

Reading the libolm data format is only required to seamlessly upgrade from libolm to vodozemac and I have working code that doesn't utilize bincode, so I can live without this.

Still, using bincode is much neater and this patch doesn't seem to introduce much complexity into bincode.

[1]: https://stackoverflow.com/questions/38047167/is-bool-guaranteed-to-be-1-byte
[PAE]: https://en.wikipedia.org/wiki/Physical_Address_Extension
[libolm]: https://gitlab.matrix.org/matrix-org/olm/
[vodozemac]: https://github.com/matrix-org/vodozemac/

VictorKoenders:
> a bool could use less than an u8

Not in bincode, we encode each bool as an `u8`.

> We're currently doing a Rust rewrite of libolm called vodozemac, it would be neat if I could use bincode to read the data that libolm produces.

I'd recommend making a wrapper struct that handles the encoding and decoding the way you want. We're not interested in being compatible with other formats.

> this patch doesn't seem to introduce much complexity into bincode

We have a lot of these requests and over time it will introduce much complexity and legacy code that may or may not be used. We'd rather not.

PR #498, opened 2022-02-02
Fuzz for compatibility with bincode v1

ghost:
This finds this input:

```rust
[20, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 252, 0, 0, 0, 0, 245]
```

Which fails because of

```rust
[fuzz_targets/roundtrip.rs:54] before = NonZeroU128(
    77990222474978957318644826112,
)
thread '<unnamed>' panicked at 'failed to deserialize: Custom("invalid value: integer `0`, expected a nonzero u128")', fuzz_targets/roundtrip.rs:55:17
```

Not sure if this is a bug, but making this as a draft to just make a note of it. I might want to move the compatibility check to a new fuzzer before this is merged.

### DIFF
--- a/fuzz/fuzz_targets/compat.rs
+++ b/fuzz/fuzz_targets/compat.rs
@@ -43,7 +43,7 @@ fuzz_target!(|data: &[u8]| {
     #[allow(deprecated)]
     let mut configv1 = bincodev1::config();
     configv1.limit(1024);
-    let bincode_v1: Result<AllTypes, _> = configv1.deserialize(&data);
+    let bincode_v1: Result<AllTypes, _> = configv1.deserialize(data);
     let bincode_v2: Result<(AllTypes, _), _> = bincode::decode_from_slice(data, config);
 
     if bincode_v1.as_ref().ok() != bincode_v2.as_ref().ok().map(|x| &x.0) {
--- a/fuzz/fuzz_targets/compat.rs
+++ b/fuzz/fuzz_targets/compat.rs
@@ -1,27 +1,22 @@
 #![no_main]
 use libfuzzer_sys::fuzz_target;
 
-use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
+use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::ffi::CString;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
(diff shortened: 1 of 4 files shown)


fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -40,12 +38,22 @@ enum AllTypes {
 }
 
 fuzz_target!(|data: &[u8]| {
-    let config = bincode::config::standard().with_limit::<1024>();
+    let config = bincode::config::legacy().with_limit::<1024>();
+    let configv1 = bincodev1::DefaultOptions::new().with_limit(1024).allow_trailing_bytes();
```
ghost: This will use varint encoding, the `legacy()` will use fixint. The true legacy config from bincode 1 is `bincodev1::Config`

ghost:
Okay, this seems like it's correct? Unless I'm missing something else with the config (endian?).

Example output is

```
Bytes:      [7, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0]
Bincode V2: CString("")
Bincode V1: Err(Custom("nul byte found in provided data at position: 0"))
```

ghost:
That specific issue should now be fixed in trunk.

fuzz/fuzz_targets/compat.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+fuzz_target!(|data: &[u8]| {
+    let config = bincode::config::legacy().with_limit::<1024>();
+    #[allow(deprecated)]
+    let mut configv1 = bincodev1::config();
+    configv1.limit(1024);
+    let result: Result<(AllTypes, _), _> = bincode::decode_from_slice(data, config);
+
+    if let Ok((before, _)) = result {
```
ghost: This doesn't look like it will catch the case where v2 fails to decode but v1 succeeds.

ghost: Okay, changed the logic to assert that they both either get the same result, or both fail.

ghost:
I'm getting 

```rust
Bytes:      [12, 0, 0, 0, 0, 100, 0, 0, 0, 100]
Bincode V1: Ok(Ipv4Addr(0.100.0.0))
Bincode V2: Err(UnexpectedEnd)
```

now.

VictorKoenders:
Fixed some compatibility issues in #503 including `Result`, `Ipv4Addr` and `Ipv6Addr`

ghost (approved):
I’ll look at a fix for the ip addr bug. But this PR looks good. We can run the fuzzer for a while.

PR #500, opened 2022-02-05
Add Membership test

ppamorim:
HI, as requested, I implemented the test required for crash that I am experiencing in my application.

Here is the demo for the crash in my application:

https://github.com/ppamorim/openraft/tree/bincode-bug/example-raft-kv

You will need to run `./test-cluster.sh`, wait it to finish and read the log `n1.log`. The code that causes the crash is here: https://github.com/ppamorim/openraft/blob/bincode-bug/example-raft-kv/src/store/mod.rs#L230-L244

The struct that causes the problem is here: https://github.com/datafuselabs/openraft/blob/main/openraft/src/membership/membership.rs#L21


compatibility/README.md
```diff
@@ -9,19 +9,19 @@ We have created a test project that can be used to test the compatibility betwee
 ## Adding a test case for your project
 
 To add a test case for your project, please follow the following steps:
-- [ ] Fork https://github.com/bincode-org/bincode
-- [ ] create a new file `compatibility/src/<name>.rs`.
-- [ ] Add a link to your project
-- [ ] Add `Licence: MIT OR Apache-2.0` if you agree to distribute your code under this license
-- [ ] Add a `mod <name>;` in the `lib.rs`. Make sure it's alphabetically ordered (check the ordering in your file system).
-- [ ] Add your structs.
+- [X] Fork https://github.com/bincode-org/bincode
```
VictorKoenders: Can you undo these changes? They're meant as a personal checklist and not to be committed

compatibility/src/membership.rs
```diff
@@ -0,0 +1,48 @@
+use std::collections::BTreeSet;
```
VictorKoenders: Missing a header like the one [here](https://github.com/bincode-org/bincode/blob/trunk/compatibility/src/rand.rs#L1-L3)

compatibility/src/membership.rs
```diff
@@ -0,0 +1,48 @@
...
+            learners: random_btreeset(&mut rng, 1000),
+            configs: vec_random_btreeset(&mut rng,),
+            all_members: random_btreeset(&mut rng, 1000),
+        });
+    }
+}
+
+fn vec_random_btreeset(rng: &mut ThreadRng) -> Vec<BTreeSet<NodeId>> {
+  let mut vec = Vec::with_capacity(5);
+  for _ in 0..10 {
```
VictorKoenders: How many cases do you have in your project? I"m wondering if we should add a random amount of btreeset's, between 0 (inclusive) and 100 maybe

compatibility/src/membership.rs
```diff
@@ -0,0 +1,48 @@
...
+    vec.push(random_btreeset(rng, 100));
+  }
+  vec
+}
+
+fn random_btreeset(rng: &mut ThreadRng, count: u32) -> BTreeSet<NodeId> {
+  let mut set = BTreeSet::new();
+  for _ in 0..count {
```
VictorKoenders: This count is static and could be random. Maybe 0-100 or something?

VictorKoenders (changes requested):

compatibility/src/membership.rs
VictorKoenders: Something like this would work:
```rs
// Added by [ppamorim](https://github.com/ppamorim)
// Taken from https://github.com/datafuselabs/openraft/blob/209ae677ade5b624fea9f6630e9ff191963f5d74/openraft/src/membership/membership.rs#L21
// License: Openraft is licensed under the terms of the MIT License or the Apache License 2.0, at your choosing.
```

VictorKoenders:
Found some compatibility problems, see #503 for more info

compatibility/src/membership.rs
ppamorim: It's a very low number of entries but I will modify that to use a random amount.

compatibility/src/membership.rs
ppamorim: Thank you

ppamorim:
I am a disaster to rebase with merged branches.

VictorKoenders (approved):

PR #507, opened 2022-02-22
Reference implementations for Reader and Writer

BRA1L0R:
I added a `Writer` implementation for `&mut T: Writer`, as there's one for Reader already.

I also removed unnecessary lifetime generics from the reference implementation for Reader since they are not needed.


VictorKoenders (approved):
Looks good to me, thanks!

PR #522, opened 2022-03-14
Added `additional` to the `UnexpectedEnd` decode error

### LINKED ISSUE #521: Change `UnexpectedEnd` to include the expected size
Right now the enum variant of ErrorKind `UnexpectedEnd` is void. I suggest changing it to `UnexpectedEnd(usize)` so the caller will be able to know how much data it is missing for success.

This will allow development of stream protocols to be more efficient by reserving buffers & calling the decode only when data is complete.

More specifically, I'm now developing a `actix-codec` `Encoder`/`Decoder` and I'd wish to reserve the additional missing data before returning.

Thoughts?

VictorKoenders:
Closes #521


ghost (approved):

aviramha:
Thank you!!!

PR #525, opened 2022-03-15
Add zoxide under Bincode in the Wild

ajeetdsouza:
Hey, [zoxide](https://github.com/ajeetdsouza/zoxide) seems to be a different use of Bincode from the ones mentioned in the README, so I thought you might like to include it here.

Since it's a shell plugin, I decided to optimize for performance as much as possible. zoxide has to store a rather small database, so Bincode becomes a really good choice -- at that size, it's significantly faster than SQLite as well as a couple of other binary formats I tested against.


VictorKoenders (approved):
LGTM

I tried porting zoxide to bincode 2.0.0-rc, but I ran into some issues with the `#[serde(borrow)]` flag you use for `Vec<Dir>`. This will require some internal rewrites, but expect a PR on that soon.

Thanks for letting us know!

ajeetdsouza:
That sounds great, thank you!

PR #526, opened 2022-03-15
Rewrite: seperated Decode and BorrowDecode

### LINKED ISSUE #511: #[bincode(bound)] for generic types
I'm looking into switching to bincode v2 and I'd love to get rid of serde - the only thing that seems to be lacking in the derive macros is that I can't specify custom bounds for generics. e.g.:

```rust
#[derive(Encode, Decode)]
struct A<B: MyTrait> {
    foo: usize,
    b: B,
    thing: B::Assoc,
}
```
With serde, I had a `#[serde(bound(deserialize = "B::Assoc: serde::Deserialize<'de>", serialize = "B::Assoc: serde::Serialize"))]`; it'd be awesome to have a similar feature for bincode v2's derive.

### LINKED ISSUE #527: `Arc<str>` can't be bincoded
`Arc<str>` can't be bincoded with bincode 2. See code and error message below. `Arc<&str>` works though. Link to gist at bottom.

```rust
use std::sync::Arc;
use bincode;

fn main() {
  let val: Arc<str> = std::sync::Arc::from("Example String");
  let config = bincode::config::standard();
  let encoded = bincode::encode_to_vec(val, config).unwrap();


  for byte in encoded {
    print!(" {byte:02x}");
  }
  println!();
}
```

compiled on rustc 1.61.0-nightly (58f11791a 2022-03-17) on MacOS gives:

```
error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> minimal_example/src/main.rs:7:40
   |
7  |   let encoded = bincode::encode_to_vec(val, config).unwrap();
   |                 ---------------------- ^^^ doesn't have a size known at compile-time
   |                 |
   |                 required by a bound introduced by this call
   |
   = help: the trait `Sized` is not implemented for `str`
   = note: required because of the requirements on the impl of `Encode` for `Arc<str>`
note: required by a bound in `encode_to_vec`
  --> /Users/nalp/.cargo/registry/src/github.com-1ecc6299db9ec823/bincode-2.0.0-rc.1/src/features/impl_alloc.rs:42:25
   |
42 | pub fn encode_to_vec<E: enc::Encode, C: Config>(val: E, config: C) -> Result<Vec<u8>, EncodeError> {
   |                         ^^^^^^^^^^^ required by this bound in `encode_to_vec`
```

https://gist.github.com/nalply/c29385f0046ca03ae0f3a687dfeb24d1

VictorKoenders:
There are a number of issues with the current design where `Decode` depends on `BorrowDecode`. The most obvious one is the implementation of `Cow`.

Ideally we'd want:

```rs
impl<T> BorrowDecode for Cow<T> where T: BorrowDecode {
    fn borrow_decode<D: BorrowDecoder<'de>>(decoder: &mut D) -> Result<Self, DecodeError> {
        Ok(Cow::Borrowed(T::borrow_decode(decoder)?))
    }
}

impl<T> Decode for Cow<T> where T: Decode {
    fn decode<D: Decoder<'de>>(decoder: &mut D) -> Result<Self, DecodeError> {
        Ok(Cow::Owned(T::decode(decoder)?))
    }
}
```

But this is currently not possible.

This issue happens with other containers types as well. One issue is `Arc<str>` (#527), because `Arc<T>` requires `T: Decode`.

This pull request splits the implementation of `Decode` and `BorrowDecode` to be completely separate.

Some notes:
- I added a `impl_borrow_decode` macro that is publicly available, which can be used to automatically derive `BorrowDecode` for any type that implements `Decode`
  - This only works with basic types. Types with generics for example cannot use this macro
- `#[derive(Decode)]` now implies `#[derive(BorrowDecode)]` automatically.

As far as I can tell this is only a breaking change for people who implement `Decode` manually. Otherwise it should hopefully be a painless change.


ghost:
Will this require anyone who implements `Decode` to also implement `BorrowDecode`? i.e. is it considered an error to implement _just_ `Decode`?

VictorKoenders:
They should implement both, but it's not enforced in the type system. I'm not sure if we should considering it an error

src/de/impl_tuples.rs
```diff
@@ -126,16 +250,607 @@ where
...
+            L::decode(decoder)?,
+            M::decode(decoder)?,
+            N::decode(decoder)?,
+            O::decode(decoder)?,
+            P::decode(decoder)?,
+        ))
+    }
+}
+
+// For re-generating the above code
```
ghost: Why don't we just keep the macro code instead of having the manual expansion?

VictorKoenders:
Some cases we should test:
- `Arc<str>`: #527
- `Cow<'a, str>`: https://github.com/ajeetdsouza/zoxide/blob/main/src/db/dir.rs#L83-L88

src/features/impl_alloc.rs
```diff
@@ -255,6 +362,15 @@ where
         Ok(vec.into_boxed_slice())
     }
 }
+// TODO
```
ghost: Why is this TODO?

src/de/impl_tuples.rs
VictorKoenders: I thought the macro would be slower than the exported code. Testing this seems to be partly true, but not enough to matter:

Build times were ~3.48 seconds with the exported code, and ~3.52 seconds with the macro.

src/features/impl_alloc.rs
VictorKoenders: Because I forgot, implemented it now

fuzz/Cargo.lock
```diff
@@ -122,5 +123,3 @@ checksum = "8ccb82d61f80a663efe1f787a51b16b5a51e3314d6ac365b08639f52387b33f3"
 [[package]]
 name = "virtue"
 version = "0.0.7"
```
ghost: This doesn't seem to have regenerated the lock file since its still pointing to virtue 0.0.7.

src/de/impls.rs
```diff
@@ -407,17 +436,17 @@ impl<'a, 'de: 'a> BorrowDecode<'de> for &'a [u8] {
...
-        match super::decode_option_variant(decoder, core::any::type_name::<Option<&[u8]>>())? {
-            Some(_) => {
-                let val = BorrowDecode::borrow_decode(decoder)?;
-                Ok(Some(val))
-            }
-            None => Ok(None),
-        }
-    }
-}
+// impl<'a, 'de: 'a> BorrowDecode<'de> for Option<&'a [u8]> {
```
ghost: You have left commented out code in the commit. Burn it :fire:

src/enc/impls.rs
```diff
@@ -283,12 +283,12 @@ impl Encode for char {
     }
 }
 
-impl Encode for &'_ [u8] {
-    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
-        super::encode_slice_len(encoder, self.len())?;
-        encoder.writer().write(self)
-    }
-}
+// impl Encode for &'_ [u8] {
```
ghost, line 286: More commented out code to burn with holy fire :fire:

ghost: Actually, put this back please. Comment out the generic T

ghost (changes requested):

src/de/impls.rs
VictorKoenders: I added a BlockedTODO for specialization, which is why I wanted to keep this code

src/enc/impls.rs
VictorKoenders: I added a BlockedTODO for specialization, which is why I wanted to keep this code

ghost (approved):

PR #529, opened 2022-03-24
Allow encoding/decoding of HashMap and HashSet with custom hash algorithms

bronsonp:
The current implementations of Encode and Decode for HashMap and HashSet are only defined for the default hash algorithm. However, these types also allow a custom hash type to be defined. At present, maps and sets using custom hashers cannot be handled by bincode. 

This pull request adds the necessary functionality to support custom hash algorithms. The main change is the use of `with_capacity_and_hasher` instead of `with_capacity` when creating the new object in the Decode implementation.

The pull request also adds test cases in `tests/std.rs` to ensure that both HashMap and HashSet with custom hash algorithms can be handled successfully.

### DIFF
--- a/src/features/impl_std.rs
+++ b/src/features/impl_std.rs
@@ -361,7 +361,7 @@ impl Decode for SocketAddrV6 {
 impl std::error::Error for EncodeError {}
 impl std::error::Error for DecodeError {}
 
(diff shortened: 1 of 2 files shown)


VictorKoenders (approved):
Looks good to me, thanks! 👍

PR #533, opened 2022-04-01
Fix riscv32 atomics and fix tests on 32-bit platforms

xobs:
This fixes tests on 32-bit platforms. Previously, the `Encode` would encode a `usize` and decode a `u64`. This works fine on 64-bit platforms, but on 32-bit platforms this results in a mismatch.

Force `usize` to be encoded as a `u64` even on 32-bit platforms.

Additionally, the 32-bit riscv target does not implement `AtomicU64` or `AtomicI64`, so remove those impelementations on that platform.

### DIFF
--- a/src/features/atomic.rs
+++ b/src/features/atomic.rs
@@ -1,8 +1,11 @@
 use crate::{de::Decode, enc::Encode};
 use core::sync::atomic::{
-    AtomicBool, AtomicI16, AtomicI32, AtomicI64, AtomicI8, AtomicIsize, AtomicU16, AtomicU32,
-    AtomicU64, AtomicU8, AtomicUsize, Ordering,
+    AtomicBool, AtomicI16, AtomicI32, AtomicI8, AtomicIsize, AtomicU16, AtomicU32, AtomicU8,
+    AtomicUsize, Ordering,
 };
+
+#[cfg(target_has_atomic = "64")]
+use core::sync::atomic::{AtomicI64, AtomicU64};
 
 impl Encode for AtomicBool {
     fn encode<E: crate::enc::Encoder>(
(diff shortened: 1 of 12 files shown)


src/features/atomic.rs
```diff
@@ -64,6 +67,7 @@ impl Decode for AtomicU32 {
     }
 }
 
+#[cfg(not(target_arch = "riscv32"))]
```
VictorKoenders: This should also be `target_has_atomic`

src/features/atomic.rs
```diff
@@ -73,6 +77,7 @@ impl Encode for AtomicU64 {
...
```
VictorKoenders: This should also be `target_has_atomic`

VictorKoenders (commented):
I wonder if we could make this bigger and add `target_has_atomic` to all atomic impls, and deprecate the `atomic` feature

xobs:
Thanks for the suggestions. I've annotated all Atomic implementations.

I assume that `Bool` is the same as `8`, and `usize` is the same as `ptr`.

I'm not sure what the best way is to deprecate the feature, so I removed it wholesale.

VictorKoenders:
Can you pull in the new cross-platform build pipeline, and see if bincode now compiles for your target arch?

xobs:
It does build for my platform still, so it looks good to me!

```sh
[10:10:39 am] E:/Code/bincode> cargo build --lib --target riscv32imac-unknown-xous-elf
   Compiling virtue v0.0.7
   Compiling bincode_derive v2.0.0-rc.1 (E:\Code\bincode\derive)
   Compiling bincode v2.0.0-rc.1 (E:\Code\bincode)
    Finished dev [unoptimized + debuginfo] target(s) in 2.39s
[10:10:43 am] E:/Code/bincode>
```

VictorKoenders:
If you run `cargo fmt` and `cargo clippy` I think this is good to merge

xobs:
Alright, I just ran `cargo fmt` to fix that up.

VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #542, opened 2022-05-09
Add impl Encode for [T], where T: Encode

cronokirby:
Since Encode takes a reference, this allows us to encode &[T] directly
using this implementation. The encoding scheme is the same as for
Vec<T>.

This also makes the implementation for &[u8] superfluous, since we get
an implementation for [u8] by virtue of u8 implementing encode. This
also gives us free implementations for &[u16], &[u32], etc. which is
quite useful. Nonetheless, we keep the implementation for &[u8] around,
because the implementation can directly write a large number of bytes,
which can be more efficient than the generic implementation.

And uh, apologies if you would have preferred I raise an issue first, but this felt like a small enough change to warrant a direct PR.

I actually ran into a situation where this was convenient. My workaround to not having this was to take `&Vec<T>` as an argument to my function, and this seemed highly unidiomatic. I was also surprised not to see this when digging around in the docs the other day.

### DIFF
--- a/src/enc/impls.rs
+++ b/src/enc/impls.rs
@@ -283,10 +283,13 @@ impl Encode for char {
     }
 }
 
+// BlockedTODO: https://github.com/rust-lang/rust/issues/37653
+//
+// We'll want to implement encoding for both &[u8] and &[T: Encode],
+// but those implementations overlap because u8 also implements Encode
 // impl Encode for &'_ [u8] {
 //     fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
-//         super::encode_slice_len(encoder, self.len())?;
-//         encoder.writer().write(self)
+//         encoder.writer().write(*self)
 //     }
 // }
 
--- a/src/enc/impls.rs
+++ b/src/enc/impls.rs
@@ -290,6 +290,19 @@ impl Encode for char {
 //     }
 // }
 
+impl<T> Encode for [T]
+where
+    T: Encode,
+{
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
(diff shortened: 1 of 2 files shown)


VictorKoenders:
I think this will conflict with #526 and I want to get that PR merged in first, so this PR will have to stay open for a while.

Thank you for your contribution regardless!

VictorKoenders:
#526 is now merged.

Looks like this PR is still valid. You can remove the implementation on `Box<[T]>` because I think the implementation on `[T]` and `Box<T>` is strictly better.

cronokirby:
Nice! I've removed redundant implementations for `Box<T>` as well as for `&[T]` (since it's made redundant since we now have both `[T]` and `&T`, and thus `&[T]`), the behavior for the latter is the same as before, so this should have no effect on end users.

src/enc/impls.rs
```diff
@@ -324,26 +337,6 @@ fn encode_utf8(writer: &mut impl Writer, c: char) -> Result<(), EncodeError> {
...
 
-// BlockedTODO: https://github.com/rust-lang/rust/issues/37653
-//
-// We'll want to implement encoding for both &[u8] and &[T: Encode],
-// but those implementations overlap because u8 also implements Encode
-// impl Encode for &'_ [u8] {
-//     fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
-//         encoder.writer().write(*self)
-//     }
-// }
```
VictorKoenders, line 335: This comment can stay, we still want to do implementation specialization

cronokirby: Maybe move the comment above the [T] implementation then?

VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #553, opened 2022-06-15
Shrink `DecodeError` from 48 to 32 bytes on 64-bit arch

maciejhirsz:
This is done by two tricks:

1. Packing the ~entire description of an enum with it's type name and~ allowed variants behind a `&'static` ref.
2. Reducing `DecodeError::CStringNulError` to just the position of the null byte.

~Aside from the extra struct being added, this is a breaking change due to type names for many manual impls being reduced to just the identifier name (derived impls were already just the identifier here). This is unfortunately necessary as `std::any::type_name` is not `const`, and even if it were using it with an external generic type would make it unusable in const environment, meaning it's impossible to construct a static `AllowedEnum` struct with it.~ `CStringNulError` I think is exotic enough and rare enough that dropping information here makes sense, if not we could also `Box` the internal error (std already allocates there anyway).

~If these changes are unacceptable, it's still possible to reduce the size of `DecodeError` to 40 bytes by just putting the `AllowedEnumVariants` behind a `&'static` ref, as that enum alone is 3 words wide.~

### DIFF
--- a/derive/src/derive_enum.rs
+++ b/derive/src/derive_enum.rs
@@ -185,7 +185,7 @@ impl DeriveEnum {
                 if self.variants.iter().any(|i| i.has_fixed_value()) {
                     // we have fixed values, implement AllowedEnumVariants::Allowed
                     variant_inner.push_parsed(format!(
-                        "{}::error::AllowedEnumVariants::Allowed",
+                        "&{}::error::AllowedEnumVariants::Allowed",
                         crate_name
                     ))?;
                     variant_inner.group(Delimiter::Parenthesis, |allowed_inner| {
@@ -204,7 +204,7 @@ impl DeriveEnum {
                 } else {
                     // no fixed values, implement a range
                     variant_inner.push_parsed(format!(
-                        "{0}::error::AllowedEnumVariants::Range {{ min: 0, max: {1} }}",
+                        "&{0}::error::AllowedEnumVariants::Range {{ min: 0, max: {1} }}",
                         crate_name,
                         self.variants.len() - 1
                     ))?;
--- a/src/de/impls.rs
+++ b/src/de/impls.rs
@@ -607,7 +607,7 @@ where
             }
             x => Err(DecodeError::UnexpectedVariant {
                 found: x as u32,
-                allowed: crate::error::AllowedEnumVariants::Range { max: 1, min: 0 },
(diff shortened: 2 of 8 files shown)


VictorKoenders:
The `const`ness of `core::any::type_name` is being tracked here: https://github.com/BlockedTODO/BlockedTODO#readme

Looks like `TypeId::new` got `const`'d but it got reverted for reasons: https://github.com/rust-lang/rust/issues/77125

maciejhirsz:
Actually I think I miscalculated, and we can get 32 bytes wide `DecodeError` by just having the `AllowedEnumVariants` behind a ref: `type_name` is 2 words, `allowed` (`AllowedEnumVariants`) behind a ref is 1 word, `found` is half a word and can fit alongside the tag -> 4 words total.

Gonna redo that.

VictorKoenders:
Can we add a test to validate the size of all `Error` types in bincode? That way we know if they change and what the actual value is

maciejhirsz:
FYI: this is the variant that makes `EncodeError` 40 bytes wide with `std` on:

```rust
    /// The encoder tried to encode a `SystemTime`, but it was before `SystemTime::UNIX_EPOCH`
    #[cfg(feature = "std")]
    InvalidSystemTime {
        /// The error that was thrown by the SystemTime
        inner: std::time::SystemTimeError,
        /// The SystemTime that caused the error
        time: std::time::SystemTime,
    },
```

Could remove `time`, or box one or the other.

VictorKoenders:
Interesting, on `x86_64-pc-windows-gnu` + `std`, the `EncodeError` is 32 bytes

But why

maciejhirsz:
> Interesting, on `x86_64-pc-windows-gnu` + `std`, the `EncodeError` is 32 bytes
> 
> But why

Kind of makes sense, `SystemTime` is platform specific and it's more precise on Linux than Windows. What is interesting is that `SystemTimeError` is a newtype for `Duration`, which has the same layout all platforms.

I've boxed `time` in that variant, that should keep the whole thing behave well on all platforms with 32-byte `EncodeError`.

VictorKoenders:
Oh I thought we were compiling against `-msvc` as well, and I was very confused that `-gnu` was different.

Boxing the SystemTime seems like a good solution to me

VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #581, opened 2022-09-22
Document configuration generics

### LINKED ISSUE #559: Fixint et. al. are doc hidden but probably should be able to lookup
The `Fixint`, `LittleEndian` types for the Configuration are marked as `doc(hidden)`, but probably shouldn't be as there is a need to find their import path sometimes:

E.g., if one declares a config as static the compiler wants the generic annotation:

```
static BINCODE_CONFIG: bincode::config::Configuration<LittleEndian, Fixint> =
    bincode::config::standard()
        .with_fixed_int_encoding()
        .with_little_endian();
```

The current docs don't allow me to lookup where LittleEndian, Fixint are declared in order to import them, unless I find them in the source (because they're hidden).

trevyn:
Closes #559

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -212,7 +212,7 @@ impl<T> Config for T where
 {
 }
 
-#[doc(hidden)]
+/// Encodes all integer types in big endian.
 #[derive(Copy, Clone)]
 pub struct BigEndian {}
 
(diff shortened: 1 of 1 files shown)


VictorKoenders (approved):
LGTM, thanks!

PR #583, opened 2022-09-27
Implement Encode for tuples with up-to 16 elements.

gz:
This makes it match the implementation of Decode which already supports up to 16 fields.

### DIFF
--- a/src/enc/impl_tuples.rs
+++ b/src/enc/impl_tuples.rs
@@ -136,3 +136,269 @@ where
         Ok(())
     }
 }
+
+impl<A, B, C, D, E, F, G, H, I> Encode for (A, B, C, D, E, F, G, H, I)
+where
+    A: Encode,
+    B: Encode,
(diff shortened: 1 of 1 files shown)


VictorKoenders:
I forget why this isn't a macro as well, but I see no reason to block this PR on that

VictorKoenders (approved):
Thanks!

VictorKoenders:
For your vmware PR; the next release candidate milestone is tracked here: https://github.com/bincode-org/bincode/milestone/6

gz:
> I forget why this isn't a macro as well, but I see no reason to block this PR on that

I tried to improve that, but struggled with generating code that uses .0, .1, .2 etc. when trying to access the tuple members A, B, C etc. for calling encode. So eventually I gave up.

For future reference: Maybe the right way is to unpack the tuple with something like `let $tuple_name($($element),*) = self;`

PR #585, opened 2022-10-01
Extended BorrowDecode for HashMap to support custom hashers

Speedy37:


VictorKoenders (approved):
I really need to get better at catching this myself.

Thanks for the PR!

PR #586, opened 2022-10-02
Allow decoding with custom `DeserializeSeed`

MrGVSV:
This allows users to decode using a custom seed. It basically just uses the same logic behind `decode_borrowed_from_slice`, but instead accepts a `DeserializeSeed<'de>`.

##### Example

```rust
let seed = MyDeserializeSeed::new(/* ... */);
let value = bincode::serde::decode_seed_from_slice(
  seed,
  input.as_slice(),
  bincode::config::standard(),
)
.unwrap();
```

### DIFF
--- a/src/features/serde/de_borrowed.rs
+++ b/src/features/serde/de_borrowed.rs
@@ -25,6 +25,28 @@ where
     T::deserialize(serde_decoder)
 }
 
+/// Decode a borrowed type from the given slice using a seed. Some parts of the decoded type are expected to be referring to the given slice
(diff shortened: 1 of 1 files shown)


VictorKoenders:
I don't see a reason not to add this. If you can update to the latest `trunk` and undo the formatting changes, this should be good to go

MrGVSV:
Okay everything should be good to go! Just ran `cargo fmt --all` to get the remaining formatting problems (probably should've done that to start lol).

VictorKoenders (approved):
Thanks!

PR #591, opened 2022-10-08
Bump `virtue` to 0.0.9 and add test for #537

### LINKED ISSUE #537: #[derive(Encode)] and friends break on structs with default generic arguments
I'm testing against 2.0.0-rc.1.

The derive macros cause syntax errors when given a struct with a default generic argument, such as:

```
#[derive(Encode)]
struct Foo<Bar=Baz> {}
```

This produces the following error:

```
error: Invalid rust syntax, expected ', > or an ident, got Some(Punct { ch: '=', spacing: Alone, span: #0 bytes(104880..104881) })
  --> src/xxx.rs:87:15
   |
87 | struct Foo<Bar=Baz> {}
```

(I tested this in some project, so there is an unhelpful line number.)

trevyn:
Fixes #537

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -19,4 +19,4 @@ description = "Implementation of #[derive(Encode, Decode)] for bincode"
 proc-macro = true
 
 [dependencies]
-virtue = "0.0.8"
+virtue = "0.0.9"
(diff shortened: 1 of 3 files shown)


VictorKoenders (approved):
Thanks!

PR #593, opened 2022-10-08
Encode variant index instead of variant value

### LINKED ISSUE #592: negative explicit variant discriminator not allowed in enums
From https://github.com/bincode-org/bincode/issues/537#issuecomment-1114170682:

> `error: Invalid rust syntax, expected literal, got Some(Punct { ch: '-', spacing: Alone, span: #0 bytes(20762..20763) })`
> 
> ```rust
> #[derive(Encode, Decode)]
> pub enum TypeOfFile {
>     Unknown = -1,
> }
> ```

It appears that enum discriminators are encoded as `u32`, and changing this to `i32` will change the byte encoding of positive discriminators as well.

I've done some work on a patch for this if this is a desirable change.

trevyn:
If this looks good, a new crates.io release of `virtue` is needed.

Closes #592

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -19,4 +19,4 @@ description = "Implementation of #[derive(Encode, Decode)] for bincode"
 proc-macro = true
 
 [dependencies]
-virtue = "0.0.9"
+virtue = "0.0.10"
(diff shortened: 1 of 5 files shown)


VictorKoenders:
Virtue 0.0.10 is released

VictorKoenders (approved):
LGTM, thanks!

PR #602, opened 2022-11-21
Add DecodeError::Other

odysa:
There is an ` EncodeError::Other(&'static str)` in
https://github.com/bincode-org/bincode/blob/ba4d19695c837dc13b492ffa8cee602d20abba0a/src/error.rs#L18-L19
but not in `DecodeError`.

I think adding `Other(&'static str)` to `DecodeError` makes errors more user-friendly.


VictorKoenders (approved):
LGTM, thanks!

PR #604, opened 2022-11-23
Add CIFuzz GitHub Action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where Bincode already runs. CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'bincode'
+        dry-run: false
+        language: rust
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'bincode'
+        fuzz-seconds: 300
+        dry-run: false
+        language: rust
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


VictorKoenders:
Thanks for this!

I'm thinking that additionally we should also run this every night for a longer time. Maybe 30 minutes (1800 seconds)? We'd need another CI step with the following trigger:

```json
{
  "name": "CIFuzz (nightly run)",
  "on": {
    "schedule": [
      {
        "cron": "0 0 * * *"
      }
    ]
  },
  // ...
```
  
Additionally can you run this through a yaml-to-json converter to match the other CI steps? There are several of them online

DavidKorczynski:
> Additionally can you run this through a yaml-to-json converter to match the other CI step

done

> I'm thinking that additionally we should also run this every night for a longer time. Maybe 30 minutes (1800 seconds)?

I wouldn't recommend this -- this is the duty of OSS-Fuzz which will run it continuously, generate an increasingly larger corpus and then CIFuzz uses this corpus in the CI action as well. Bincode was integrated into OSS-Fuzz here: https://github.com/google/oss-fuzz/pull/8045 and you can find the relevant files for the project here: https://github.com/google/oss-fuzz/tree/master/projects/bincode -- it looks like @nmccarty was added as a primary contact in the project configuration, should more be added?

VictorKoenders:
Both nmccarty  and I were unaware that bincode was already in google's open source fuzzing system. Zoey knew about it but only because someone else pointed this out.

I guess it's a good thing we never knew that bincode was actively being fuzzed because it means we're doing something right.

> it looks like @nmccarty was added as a primary contact in the project configuration, should more be added?

Keeping nmccarty as a primary contact is fine, thanks!

VictorKoenders (approved):
LGTM, thanks!

Issue #539, opened 2022-04-09
[Tracking issue] Querying number of bytes needed to encode a struct


bitwiseshiftleft:
I've noticed that v2.0 doesn't have a serialized size function.  I'm a rust newbie, so I'm not going to attempt a pull request, but an implementation like this would probably suffice?

```
use bincode::{Encode,config::Config,enc::EncoderImpl,error::EncodeError,enc::write::Writer};

/** Writer which only counts the bytes "written" to it */
struct SizeOnlyWriter<'a> {
    bytes_written: &'a mut usize
}

impl<'a> Writer for SizeOnlyWriter<'a> {
    fn write(&mut self, bytes: &[u8]) -> Result<(), EncodeError> {
        *self.bytes_written += bytes.len();
        Ok(())
    }
}

/** Return the serialized size of an `Encode` object. */
pub fn serialized_size<T:Encode,C:Config>(obj:&T, config:C) -> Result<usize, EncodeError> {
    let mut size = 0usize;
    let writer = SizeOnlyWriter { bytes_written: &mut size };
    let mut ei = EncoderImpl::new(writer, config);
    obj.encode(&mut ei)?;
    Ok(size)
}
```

VictorKoenders:
We haven't implemented `serialized_size` because we're not sure it has a valid use case. Most of the time when people use `serialized_size` it is because they want to use it like this:

```rs
let size = bincode::serialize_size(&T, &config).unwrap();
let mut vec = vec![0u8; size];
bincode::encode_into_slice(&T, vec.as_mut_slice(), config).unwrap();
```

We personally have found that the above is a lot slower than simply encoding to a `vec`, as you have to process the entire structure twice.

```rs
let mut vec = bincode::encode_to_vec(&T, config).unwrap();
```

But maybe there's a use case we missed?

bitwiseshiftleft:
I was thinking of using it in an FFI call, so that I could serialize into memory owned by C.

VictorKoenders:
Would it be possible to:
- let bincode allocate a `Vec<u8>`
- call [into_raw_parts](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.into_raw_parts)
- return that to C
- ffi back into Rust to deallocate this vec when you're done

[...]

PR #609, opened 2023-01-24
Add EncodedSize trait to calculate encoded sizes

markbt:
Related issue: #539 

Hi bincode developers.  This is the first in a series of 3 PRs that would allow us to replace our current use of [Abomonation](https://github.com/TimelyDataflow/abomonation/) with bincode, without a noticeable performance hit.  We'd like these to be adopted upstream, and hope they are useful for other users of bincode also.

For some storage back-ends it's necessary to know the size of the buffer to pre-allocate before encoding.  This is true of our cache back-end, which is C++ code that we communicate with over FFI.  The bincode 1.0 version had `bincode::serialized_size`, which would tell you the serialized size of data, but for bincode 2.0 this is not implemented.

One suggestion is to implement a `SizeOnlyWriter` which adds up the sizes of the slices it's been given, however this effectively requires serializing the data twice, as it doesn't allow types to efficiently compute the size when it is possible to do so without performing the actual serialization steps.

In this PR, we add a new `EncodedSize` trait which accomplishes this objective, and implement that for all types.  Implementations for custom structs and enums can be derived using `#[derive(bincode::EncodedSize)]`.

### DIFF
--- a/derive/src/attribute.rs
+++ b/derive/src/attribute.rs
@@ -7,6 +7,7 @@ pub struct ContainerAttributes {
     pub decode_bounds: Option<(String, Literal)>,
     pub borrow_decode_bounds: Option<(String, Literal)>,
     pub encode_bounds: Option<(String, Literal)>,
+    pub encoded_size_bounds: Option<(String, Literal)>,
 }
 
 impl Default for ContainerAttributes {
@@ -17,6 +18,7 @@ impl Default for ContainerAttributes {
             decode_bounds: None,
             encode_bounds: None,
             borrow_decode_bounds: None,
+            encoded_size_bounds: None,
         }
     }
 }
@@ -71,6 +73,15 @@ impl FromAttribute for ContainerAttributes {
                     let val_string = val.to_string();
                     if val_string.starts_with('"') && val_string.ends_with('"') {
                         result.borrow_decode_bounds =
+                            Some((val_string[1..val_string.len() - 1].to_string(), val));
+                    } else {
+                        return Err(Error::custom_at("Should be a literal str", val.span()));
+                    }
+                }
+                ParsedAttribute::Property(key, val) if key.to_string() == "encoded_size_bounds" => {
+                    let val_string = val.to_string();
+                    if val_string.starts_with('"') && val_string.ends_with('"') {
+                        result.encoded_size_bounds =
                             Some((val_string[1..val_string.len() - 1].to_string(), val));
                     } else {
                         return Err(Error::custom_at("Should be a literal str", val.span()));
--- a/derive/src/derive_enum.rs
+++ b/derive/src/derive_enum.rs
@@ -135,6 +135,122 @@ impl DeriveEnum {
         Ok(())
     }
 
+    pub fn generate_encoded_size(self, generator: &mut Generator) -> Result<()> {
+        let crate_name = self.attributes.crate_name.as_str();
+        generator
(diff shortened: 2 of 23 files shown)


VictorKoenders:
Hi, and thank you for this PR!

Scanning over this PR, I'm not actually sure this will be faster than `SizeOnlyWriter`. This PR still has to iterate over all the fields recursively and count all the entries.

Is it possible to set up a benchmark to compare the speed of an `SizeOnlyWriter` implementation, and this PR?

markbt:
Benchmarking is a good idea.  It ended up being a bit of a journey, and it looks like you're right, as long as LTO is enabled.

To start with, I added a variant of `encoded_size` to `bincode` which used `SizeOnlyWriter`.  I do think it would be good to provide this as part of the library, if only as a convenience, to save everyone re-implementing it.  I created a benchmark that computes the serialized size a `Vec` of 10,000 random integers of various sizes.

Results were initially pretty promising for the `EncodedSize` trait (all times in nanoseconds):

| | u8 | u16 | u32 | u64 |
|--|--:|--:|--:|--:|
| SizeOnlyWriter | 14770.00 | 29581 | 29416 | 37906 |
| **EncodedSize** | 1.45 | 12686 | 10511 | 12582 |

Yes, for `Vec<u8>` it's 10,000 times faster with `EncodedSize`.

That seemed pretty obvious to be an optimization problem, with `Vec<u8>` likely not being optimized to `len(x)` as it was with the `EncodedSize` variant of `Vec<u8>`.

To test this out, I moved the `SizeOnlyWriter` implementation to the benchmark crate:

| | u8 | u16 | u32 | u64 |
|--|--:|--:|--:|--:|
| Local SizeOnlyWriter | 0.206 | 3953 | 5741 | 7257 |

That's *much* faster, and even beats out the dedicated trait.  I don't think this is a general-purpose solution: I don't think it's reasonable to copy and paste `SizeOnlyWriter` to every location in your codebase that you want to compute the size.  Also, it's likely that it won't scale well, as it'll mean a new monomorphisation of the same code each time.  But it gives a good lower bound.

Then I realised that LTO wasn't enabled, and that it might help.  It did:

| | u8 | u16 | u32 | u64 |
|--|--:|--:|--:|--:|
| SizeOnlyWriter + LTO | 0.206 | 3864 | 5671 | 7218 |
| EncodedSize + LTO | 0.206 | 3876 | 5707 | 7268 |
| Local SizeOnlyWriter + LTO | 0.206 | 3877 | 5676 | 7242 |

So ultimately I think using `SizeOnlyWriter` is probably fine.  I do think it would be nice if this was built into `bincode` as the recommended way of doing this (with the caveat that you shouldn't bother unless you really need to, of course).

Will leave this PR up for a bit in case anyone disagrees, but I now plan to close this PR.

The benchmark is in [this commit](https://github.com/markbt/bincode/commit/7a98fad01dd7fade3a24861c7794395f5df76e09) if anyone wants to try it out themselves.

VictorKoenders:
> I do think it would be nice if this was built into bincode

I feel like added functionality like this this would be better in a different crate. In https://github.com/bincode-org/bincode/issues/578#issuecomment-1266714493 I talked about introducing a crate named `bincode-contrib` or `bincode-utils` for common utilities that can be build on top of bincode without big changes.

Unfortunately both Zoey and I have been very busy the last couple of months so we have not had time to set this up yet

PR #612, opened 2023-01-31
Implement support for smallvec

markbt:
Hi bincode developers. This is the second in a series of 3 PRs that would allow us to replace our current use of [Abomonation](https://github.com/TimelyDataflow/abomonation/) with bincode, without a noticeable performance hit. We'd like these to be adopted upstream, and hope they are useful for other users of bincode also.

We use `SmallVec` in a number of places in the structures that we serialize, as this gives us a noticeable memory footprint and performance improvement.  Rust's orphan rule prevents us from implementing the bincode traits for `SmallVec` ourselves, as it's defined in a different crate.  Instead, we propose adding a new optional feature to bincode with implements the traits for that type when enabled.  When the `smallvec` feature is specified, bincode additionally provides implementations of `Encode` and `Decode` for `SmallVec`, much like it already does for the standard library types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,10 +29,12 @@ default = ["std", "derive"]
 std = ["alloc", "serde?/std"]
 alloc = ["serde?/alloc"]
 derive = ["bincode_derive"]
+smallvec = ["dep:smallvec"]
 
 [dependencies]
 bincode_derive = { path = "derive", version = "2.0.0-rc.2", optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
+smallvec = { version = "1.0", optional = true }
 
 # Used for tests
 [dev-dependencies]
--- a/src/features/impl_smallvec.rs
+++ b/src/features/impl_smallvec.rs
@@ -1,0 +1,63 @@
+use crate::de::BorrowDecoder;
+use crate::de::Decode;
+use crate::de::Decoder;
+use crate::enc::Encode;
+use crate::enc::Encoder;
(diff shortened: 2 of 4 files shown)


VictorKoenders:
Ideally we would want to have other crates implement bincode, instead of bincode implementing all of the crates in the rust ecosystem.

Have you tried using `#[bincode(with_serde)]` as described [here](https://docs.rs/bincode/2.0.0-rc.2/bincode/migration_guide/index.html#bincode-derive-and-libraries)? I wonder if that will give an ergonomic-enough API for you to work with.

We'll run into issues when we finally release bincode 2.0 and other crates are thinking of implementing bincode in their crate, but bincode also provides an implementation. I think if we are going to accept features like this, they should be obviously marked that they can be removed at any time, e.g. `smallvec-unstable`. We'll reserve the right to remove those at any point in the future.

markbt:
Using `#[bincode(with_serde)]` is about a 5-10% perf regression, so I'd rather not do that if it can be avoided.  If the preferred mode is for other crates to implement your traits, then I will see if the smallvec developers will accept a small PR to add support for `bincode` there.

markbt:
Smallvec will accept my change there once bincode 2.0 is released, so I will close this PR shortly.  Do you know when a 2.0 release is likely?

VictorKoenders:
Unfortunately both ZoeyR and I have been very busy so it might be a while. We'll need to fix all the issues labeled as "bincode 2" and then I think we should be able to launch

Issue #618, opened 2023-02-18
Bincode 2 performance tracking


JojiiOfficial:
I was curious whether v2 brings performance improvements as it doesn't rely on serde. So I decided to do a quick criterion benchmark and was kinda surprised when I saw the results. 
As you can see on the screenshot below encoding is around 100x slower than v1 and decoding around 5-8x slower. I'm personally using bincode over other formats as I don't need human readable (de)serialization of my data but rather want it to be as fast as possible as my entire applications performance depends on it. So it would be nice if v2 doesn't have these performance drawbacks. Of course one could continue use the serde implementation (although I haven't tested the performance of serde in v2) but such a performance drawback in a crate like bincode should be considered as bug in my opinion, so I created this issue.
You can find the code here: https://github.com/JojiiOfficial/bincode_bench

![image](https://user-images.githubusercontent.com/15957865/219901127-8039c5ed-9790-4dd4-8e36-8f2778e748d4.png)

VictorKoenders:
Turning on `[profile.bench] lto = true` in the `Cargo.toml` massively increases the performance. In initial testing we found bincode-2 to be 10-20% faster, but it looks like this is no longer the case

![image](https://user-images.githubusercontent.com/2743142/219966858-22cc741f-a9d9-4efb-9e87-eb263fdaf590.png)

My guess is that we should add more `[inline]` attributes to functions

VictorKoenders:
Bincode 1 essentially does `&[u8].to_vec()`: https://github.com/bincode-org/bincode/blob/v1.x/src/de/read.rs#L126-L128

Bincode 2 aims to be usable in `no_std`, so it only provides an interface to fill a `&mut [u8]`: https://github.com/bincode-org/bincode/blob/trunk/src/de/read.rs#L19

Because of this, in bincode 2 we have to do
```rs
let mut vec = vec![0u8; len];
reader.read(&mut vec); // essentially `vec.copy_from_slice(&[u8]);`
```

Where in bincode 1 we could do
```rs
let vec = reader.read_vec()?; // essentially `&[u8].to_vec()`
```

Which is why bincode 1 is 20% faster in this specific case.

Funnily enough after https://github.com/bincode-org/bincode/pull/667 reading `Vec<u8>` is about 5 times faster in bincode 2, because we can optimize `Vec<T> where T: u8` to read the entire buffer instead of byte-by-byte. But in bincode 1 the string one is slightly more optimized.

[...]

PR #619, opened 2023-02-20
Improved encoding and decoding speed of Vec<u8>

VictorKoenders:
Running against the benchmarks in #618 

```
bench v1                time:   [48.663 µs 49.119 µs 49.573 µs]
                        change: [-0.9286% +0.8850% +2.7884%] (p = 0.35 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe

bench v2 (standard)     time:   [642.39 µs 646.82 µs 651.41 µs]
                        change: [-71.758% -71.415% -71.093%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

bench v2 (legacy)       time:   [650.62 µs 653.72 µs 656.93 µs]
                        change: [-71.322% -71.089% -70.844%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

bench v1 decode         time:   [301.58 µs 302.76 µs 303.97 µs]
                        change: [-42.809% -40.022% -36.789%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe

bench v2 decode (legacy)
                        time:   [326.89 µs 328.24 µs 329.63 µs]
                        change: [-81.836% -81.650% -81.457%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) low mild
  2 (2.00%) high mild
```

decoding seems to be on-par with bincode 1 now. Encoding is down to being a factor 10 slower.

### DIFF
--- a/src/enc/impls.rs
+++ b/src/enc/impls.rs
@@ -295,10 +295,17 @@ impl Encode for char {
 
 impl<T> Encode for [T]
 where
-    T: Encode,
+    T: Encode + 'static,
 {
     fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
         super::encode_slice_len(encoder, self.len())?;
+
+        if core::any::TypeId::of::<T>() == core::any::TypeId::of::<u8>() {
+            let t: &[u8] = unsafe { core::mem::transmute(self) };
+            encoder.writer().write(t)?;
+            return Ok(());
+        }
+
         for item in self {
             item.encode(encoder)?;
         }
--- a/src/features/impl_alloc.rs
+++ b/src/features/impl_alloc.rs
@@ -1,6 +1,10 @@
 use crate::{
     de::{read::Reader, BorrowDecoder, Decode, Decoder},
-    enc::{self, write::Writer, Encode, Encoder},
+    enc::{
+        self,
+        write::{SizeWriter, Writer},
+        Encode, Encoder,
+    },
     error::{DecodeError, EncodeError},
     impl_borrow_decode, BorrowDecode, Config,
 };
@@ -21,6 +25,12 @@ pub(crate) struct VecWriter {
 }
 
 impl VecWriter {
+    /// Create a new vec writer with the given capacity
+    pub fn with_capacity(cap: usize) -> Self {
+        Self {
+            inner: Vec::with_capacity(cap),
+        }
+    }
     // May not be used in all feature combinations
     #[allow(dead_code)]
     pub(crate) fn collect(self) -> Vec<u8> {
@@ -29,6 +39,7 @@ impl VecWriter {
 }
 
 impl enc::write::Writer for VecWriter {
+    #[inline(always)]
     fn write(&mut self, bytes: &[u8]) -> Result<(), EncodeError> {
         self.inner.extend_from_slice(bytes);
         Ok(())
@@ -40,7 +51,12 @@ impl enc::write::Writer for VecWriter {
 /// [config]: config/index.html
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
 pub fn encode_to_vec<E: enc::Encode, C: Config>(val: E, config: C) -> Result<Vec<u8>, EncodeError> {
-    let writer = VecWriter::default();
+    let size = {
+        let mut size_writer = enc::EncoderImpl::<_, C>::new(SizeWriter::default(), config);
+        val.encode(&mut size_writer)?;
+        size_writer.into_writer().bytes_written
+    };
+    let writer = VecWriter::with_capacity(size);
     let mut encoder = enc::EncoderImpl::<_, C>::new(writer, config);
     val.encode(&mut encoder)?;
     Ok(encoder.into_writer().inner)
@@ -274,7 +290,7 @@ where
             vec.resize(len, 0u8);
             decoder.reader().read(&mut vec)?;
             // Safety: Vec<T> is Vec<u8>
-            return Ok(unsafe { std::mem::transmute(vec) });
+            return Ok(unsafe { core::mem::transmute(vec) });
         }
         decoder.claim_container_read::<T>(len)?;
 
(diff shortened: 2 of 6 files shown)


VictorKoenders:
Pre-computing the size drops the execution speed down by another 50%

```
bench v1                time:   [51.224 µs 51.781 µs 52.294 µs]
                        change: [+5.4681% +7.4038% +9.4670%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 24 outliers among 100 measurements (24.00%)
  3 (3.00%) low severe
  10 (10.00%) low mild
  9 (9.00%) high mild
  2 (2.00%) high severe

bench v2 (standard)     time:   [363.31 µs 364.96 µs 366.79 µs]
                        change: [-83.761% -83.543% -83.335%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  1 (1.00%) high severe

bench v2 (legacy)       time:   [326.16 µs 327.39 µs 328.81 µs]
                        change: [-85.568% -85.459% -85.338%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe

bench v1 decode         time:   [339.01 µs 341.76 µs 344.70 µs]
                        change: [-34.888% -31.648% -27.911%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  5 (5.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe

bench v2 decode (legacy)
                        time:   [351.59 µs 357.93 µs 364.61 µs]
                        change: [-80.044% -79.714% -79.330%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe
  ```

**Comment by VictorKoenders** [comment]
I'm willing to call this good if this is okay with you @JojiiOfficial 

Execution speeds in the order of 51.781 µs and 364.96 µs are essentially instant. We can probably optimize this further, but we're hunting very specific LLVM optimizations at this point. In a real world scenario we would almost certainly not notice a difference between the two.

I'd suggest we say that this is fast enough for now and if we find a real world example where bincode 2 is significantly slower than bincode 1 we can add more benchmarks.

File: src/features/impl_alloc.rs [hunk 1]
```diff
@@ -262,10 +278,20 @@ where
 
 impl<T> Decode for Vec<T>
 where
-    T: Decode,
+    T: Decode + 'static,
```
adamreichold, line 281: Doesn't this bound rule out zero-copy deserialization of types like `Vec<&'de str>`?

JojiiOfficial:
Encoding still is 6 times slower than in v1. A real world example could be a lazy loaded data structure (eg. an index) that keeps its data encoded and if an item is accessed, it gets decoded. If you're doing this a million of times lets say you're building the index, v2 will take 6 times longer than v1 (As you do a lot of encoding when building). An index built with V2 will take 6 minutes where the same data structure will only take 1 minute if it was using bincode v1. If v1 takes 1h, v2 will take 6 hours and so on...

Not saying this fix isn't ok just want to point out that there are real world scenarios where this will probably hurt. In my opinion opening a tracking issue and merging this PR should be fine for now but having a de/serializing library for binary data that is implemented as fast as possible with as much optimizations as possible would be really nice. Maybe there are even more possible optimizations that would make it even faster than v1 🤔

VictorKoenders:
If you have a benchmark with data that lasts a minute to encode/decode that would be very interesting to dive into

JojiiOfficial:
Change the parameter for `build_data` from 100 to 2500 and it'll take around a minute with this PR merged for 100 iterations

JojiiOfficial:
The following code improves the performance even more for Strings and [u8] arrays:


```rust
impl<T> Encode for [T]
where
    T: Encode + 'static,
{
    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
        super::encode_slice_len(encoder, self.len())?;

        if core::any::TypeId::of::<T>() == core::any::TypeId::of::<u8>() {
            let t: &[u8] = unsafe { transmute(self) };
            encoder.writer().write(t)?;
            return Ok(());
        }

        for item in self {
            item.encode(encoder)?;
        }
        Ok(())
    }
}
```

image

VictorKoenders:
Added a couple of `#[inline]` and now the performance is almost identical
```
bench v1                time:   [58.788 µs 59.134 µs 59.497 µs]                                                                          
bench v2 (standard)     time:   [61.595 µs 62.437 µs 63.531 µs]                                                  
bench v2 (legacy)       time:   [62.514 µs 63.413 µs 64.394 µs]

bench v1 decode         time:   [314.30 µs 317.03 µs 319.98 µs]
bench v2 decode (legacy)
                        time:   [376.05 µs 378.22 µs 380.79 µs]
  ```

File: src/features/impl_alloc.rs (hunk 1 above)
Comment (adamreichold) [reply src/features/impl_alloc.rs:281]: To answer my own question: No, as this case goes though `impl BorrowDecode for Vec<T>`.

**Comment by JojiiOfficial** [comment]
Indeed, very nice!

File: src/enc/impls.rs [hunk 2]
```diff
@@ -295,10 +295,17 @@ impl Encode for char {
 
 impl<T> Encode for [T]
 where
-    T: Encode,
+    T: Encode + 'static,
```
stevenliebregt, line 298: This change now gives me errors when trying to derive `Encode` for types with `Vec`s that contain references.

For example:

```rust
struct MyStruct<'a> {
  some_field: Vec<&'a str>
}
```

`Argument requires that 'a must outlive 'static`

VictorKoenders: Reverted this PR in #663 and tried a better fix in #667

PR #625, opened 2023-03-30
Made arrays never encode their length

### LINKED ISSUE #613: bincode::encode_to_vec is incompatible with bincode::serde::encode_to_vec in standard configuration
Using version `2.0.0-rc.2`, the following code:

```
fn main() {
    let array: [u8; 3] = [11, 12, 13];
    let stdcfg = bincode::config::standard();
    assert_eq!(
        bincode::encode_to_vec(&array, stdcfg).unwrap(),
        bincode::serde::encode_to_vec(&array, stdcfg).unwrap(),
    );
}
```

Fails with:

```text
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[3, 11, 12, 13]`,
 right: `[11, 12, 13]`', src/main.rs:6:5
```

This means [`bincode::encode_to_vec`](https://docs.rs/bincode/2.0.0-rc.2/bincode/fn.encode_to_vec.html) isn't compatible with [`bincode::serde::encode_to_vec`](https://docs.rs/bincode/2.0.0-rc.2/bincode/serde/fn.encode_to_vec.html). This is highly surprising.

See also: [How can bincode distinguish between tuples and (fixed size) arrays?](https://users.rust-lang.org/t/how-can-bincode-distinguish-between-tuples-and-fixed-size-arrays/88756) on URLO.

VictorKoenders:
Closes #613 

Turns out serde never encodes a fixed array length with 32 elements or less. If an array is 33 elements or more, it silently falls back to the `&[T]` implementation (which does encode the length). Our test cases were only tested with large arrays.

Because the config flags `write_fixed_array_length` and `skip_fixed_array_length` were based on a wrong assumption on how serde/bincode 1 worked, these flags have been removed. They may be re-added in the future.

### DIFF
--- a/compatibility/src/lib.rs
+++ b/compatibility/src/lib.rs
@@ -22,18 +22,22 @@ where
     // This is what bincode 1 serializes to. This will be our comparison value.
     let encoded = bincode_1_options.serialize(t).unwrap();
 
-    println!("Encoded {:?} as {:?}", t, encoded);
+    println!("Encoded {t:?} as {encoded:?}");
 
     // Test bincode 2 encode
     let bincode_2_output = bincode_2::encode_to_vec(t, bincode_2_config).unwrap();
-    assert_eq!(encoded, bincode_2_output, "{:?} serializes differently", t);
+    assert_eq!(
+        encoded,
+        bincode_2_output,
+        "{t:?} serializes differently\nbincode 2 config {:?}",
+        core::any::type_name::<C>(),
+    );
 
     // Test bincode 2 serde serialize
     let bincode_2_serde_output = bincode_2::serde::encode_to_vec(t, bincode_2_config).unwrap();
     assert_eq!(
         encoded, bincode_2_serde_output,
-        "{:?} serializes differently",
-        t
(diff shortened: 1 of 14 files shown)


TommyLike:
@VictorKoenders hello, I use bincode in my project as below:
```
 signed.write_all(&bincode::encode_to_vec(
            &sig_struct,
            config::standard()
                .skip_fixed_array_length()
                .with_fixed_int_encoding()
                .with_big_endian(),
        )?)?;
```
it works well for the version of `bincode = "2.0.0-rc.2"` but when upgrade into rc.3 I found the `skip_fixed_array_length` has been removed, so what should I do for now, How to remove the length when serializing?

VictorKoenders:
@TommyLike currently there is no such option. Bincode 1 and 2 should work as you'd expect without the configuration. Is there a particular issue you're running into?

TommyLike:
> @TommyLike currently there is no such option. Bincode 1 and 2 should work as you'd expect without the configuration. Is there a particular issue you're running into?

No, I realized it works as I expected, thanks!

PR #630, opened 2023-04-07
Fix typos in Spec.md enum example

bigbass1997:
In the Enums section, there are examples using the second and third variants of an enum, but the comment incorrectly refers to them as "first variant".

Unless perhaps the intention here was to say "the variant is first"?


VictorKoenders (approved):
Looks like it was indeed a copy-paste issue, thanks!

PR #635, opened 2023-05-07
Allow generics in impl_borrow_decode

dullbananas:


VictorKoenders:
Thanks! Could you add some tests for this as well?

I think adding this implementation for some of the `std` types, I'm thinking of `Vec`, `HashMap`, etc, would also suffice as a test case

dullbananas:
Done

VictorKoenders:
Thanks!

PR #653, opened 2023-06-20
embedded-io compatibility

### LINKED ISSUE #652: Feature request: embedded-io compatibility
Hi!

It seems it is possible to use implementations of the std Read/Write traits as data sources/destinations. If there isn't anything std-specific in their implementations, do you think bincode could work with their [embedded-io](https://docs.rs/embedded-io/latest/embedded_io/blocking/index.html) counterparts?

bugadani:
Resolves #652

I'm opening this PR early because I have a few open questions:
 - embedded-io traits have an associated error type for IO errors. How should we incorporate this?
   - we can make EncodeError/DecodeError generic, and default to std::io::Error
   - we can introduce a custom io error type, map known std errors to it, and map embedded-io errors to EOF (ReadExact error) or Other.
 - `decode_from_embedded_io_read` is a bit verbose. I think I'd try to create a single encode/decode function and abstract the std/embedded options with a bincode-provided marker trait. The implementation would be more complex, but the API would be simpler. What is your preference here?

Right now we can't have both std and embedded-io features active, because of the IO error question. I intend this as a temporary restriction and plan on lifting it if the resolution of that question allows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,12 +27,14 @@ edition = "2021"
 [features]
 default = ["std", "derive"]
 std = ["alloc", "serde?/std"]
+embedded-io = ["dep:embedded-io"]
 alloc = ["serde?/alloc"]
 derive = ["bincode_derive"]
 
 [dependencies]
 bincode_derive = { path = "derive", version = "2.0.0-rc.3", optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
+embedded-io = { version = "0.4", optional = true }
 
 # Used for tests
 [dev-dependencies]
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,7 +7,7 @@ pub enum EncodeError {
     /// The writer ran out of storage.
     UnexpectedEnd,
 
-    /// The RefCell<T> is already borrowed
+    /// The `RefCell<T>` is already borrowed
     RefCellAlreadyBorrowed {
         /// The inner borrow error
         inner: core::cell::BorrowError,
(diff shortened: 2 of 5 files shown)


VictorKoenders:
Thanks for this!

> embedded-io traits have an associated error type for IO errors.

I think make an `EmbeddedIo { ... }` variant, and put that behind the config flag. They're completely different errors from std's `Io` errors. In hindsight we maybe should've named the existing variant `StdIo`.

The error type is marked as `#[non_exhaustive]` so people cannot match on all variants, so I believe it is fine to add variants based on features being active

> decode_from_embedded_io_read is a bit verbose.

What about `bincode::embedded_io::decode_from_read` or `bincode::embedded_io::decode`?

VictorKoenders:
Also please add a line to the features table in [src/lib.rs](https://github.com/bincode-org/bincode/blob/trunk/src/lib.rs#L24)

bugadani:
> I think make an EmbeddedIo { ... } variant, and put that behind the config flag. They're completely different errors from std's Io errors

That still requires the error enums to be generic over the particular embedded-io implementation, I think. Also, a particular encode/decode call will never return an error that can be both std _and_ embedded io error at the same time, so two separate variants seem a bit unnecessary to me.

I'm considering sketching up my "make errors generic" idea and we can throw it away if you don't like where it ends up. I'm usually pretty bad when I want to communicate ideas, maybe an example will make it clearer.

> What about bincode::embedded_io::decode_from_read or bincode::embedded_io::decode?

I was following the structure `std` used. That means glob reexports, which means either the names collide, or that embedded uses a different naming convention which isn't ideal UX.

bugadani:
Current implementation ignores the actual embedded-io error for simplicity. Trying to make Decode/EncodeError generic is at this point not practical.

I have also fixed a few typos I found.

Issue #648, opened 2023-06-13
BorrowCompat & Compat should impl Debug & Display


elpiel:

VictorKoenders:
Closed in #670

PR #670, opened 2023-10-14
Compat and BorrowCompat Debug and Display implementations

aegroto:
Solves #648. This is my first PR in this project so I am sorry for something does not respect the contribution standards. Please let me know if something has to be changed.

### DIFF
--- a/src/features/serde/mod.rs
+++ b/src/features/serde/mod.rs
@@ -59,6 +59,8 @@ mod de_borrowed;
 mod de_owned;
 mod ser;
 
+use core::fmt::Debug;
+
 pub use self::de_borrowed::*;
 pub use self::de_owned::*;
 pub use self::ser::*;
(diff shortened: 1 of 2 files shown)


src/features/serde/mod.rs
```diff
@@ -216,6 +218,24 @@ where
     }
 }
 
+impl<T> std::fmt::Debug for Compat<T>
```
VictorKoenders: I think this can be a simple `#[derive(Debug)]`, but I'm not too bothered by it. It should be `core::fmt::Debug` though

src/features/serde/mod.rs
```diff
@@ -216,6 +218,24 @@ where
...
+where
+    T: std::fmt::Debug,
+{
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_tuple("Compat").field(&self.0).finish()
+    }
+}
+
+impl<T> std::fmt::Display for Compat<T>
```
VictorKoenders: This should be `core::fmt::Display`

src/features/serde/mod.rs
```diff
@@ -252,3 +272,21 @@ where
         Ok(())
     }
 }
+
+impl<T> std::fmt::Debug for BorrowCompat<T>
```
VictorKoenders: This should be `core::fmt::Debug`

src/features/serde/mod.rs
```diff
@@ -252,3 +272,21 @@ where
...
+where
+    T: std::fmt::Debug,
+{
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_tuple("BorrowCompat").field(&self.0).finish()
+    }
+}
+
+impl<T> std::fmt::Display for BorrowCompat<T>
```
VictorKoenders: This should be `core::fmt::Display`

src/features/serde/mod.rs
```diff
@@ -59,6 +59,8 @@ mod de_borrowed;
 mod de_owned;
 mod ser;
 
+use core::fmt::Debug;
```
VictorKoenders: I don't think this import is used because you reference all the implementations directly.

If you want to keep this import, please put it below the `pub use`

VictorKoenders (changes requested):

aegroto:
@VictorKoenders thanks for your review, I have fixed the things you have pointed out. Let me know if anything else may be done.

VictorKoenders (approved):

VictorKoenders:
Thanks!

PR #681, opened 2023-11-12
Add getters for current configuration values

### LINKED ISSUE #598: Exposing config to implement custom Encode/Decode
I am trying to create custom Encode/Decode implementations for types in other crates that I'd like to use. However, I cant access the configuration as done on the impls.rs file of the encoder. How can I access the config (endianess, fixed/varint, etc) to write the data from the external type according to the config of the encoder/decoder?

shahn:
This was inspired by suggestions provided in #598, thanks a lot to @VictorKoenders.

Fixes #598.


VictorKoenders:
Sorry for the late response, I was going to think about this over the weekend and then Factorio happened and now it's a month later.

A couple of things we've been thinking about:
- Not sure what `derive` attributes we want on the enums, `PartialEq, Eq` is maybe not needed, `Clone` or `Debug` might be? We're not sure
- It would be nice if we could have `const fn` in traits. We were thinking of adding `#[inline(always)]` to the implementation functions but I don't want to do that before we actually benchmark that it is faster/slower, and I think that's outside of the scope of this PR.

Other than that it looks good, thanks for the contribution!

shahn:
Thanks for the merge, I think PartialEq and Eq are needed to ensure you can actually do something with the config values. Debug could be nice, but is not required to do something useful. Or perhaps I misunderstood your question?

VictorKoenders:
You understood my question correctly.

> I think PartialEq and Eq are needed to ensure you can actually do something with the config values

You can still match on the variants. This is what we do in bincode currently, the following code will still work without `PartialEq, Eq`:

```rs
                Ok(match D::C::ENDIAN {
                    Endianness::Little => u16::from_le_bytes(bytes),
                    Endianness::Big => u16::from_be_bytes(bytes),
                })
```

But this will not:

```rs
if D::C::ENDIAN ==  Endianness::Little {
    panic!("Only big endian supported");
}
```

But I'm also not sure if there's a downside to adding this

PR #684, opened 2023-12-13
Use const functions where possible

richardpringle:
I just used clippy for this. I actually assume that you will not want the brunt of this change but I figured it was easier to make the change everywhere then simply pick the ones you deem acceptable (if any).

I definitey think some of the constructors should be `const`. 

Thanks in advance!

### DIFF
--- a/src/de/decoder.rs
+++ b/src/de/decoder.rs
@@ -28,7 +28,7 @@ pub struct DecoderImpl<R, C: Config> {
 
 impl<R: Reader, C: Config> DecoderImpl<R, C> {
     /// Construct a new Decoder
-    pub fn new(reader: R, config: C) -> DecoderImpl<R, C> {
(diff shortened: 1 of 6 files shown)


VictorKoenders:
Looks good to me.

Out of pure interest, can you check if this actually improves any of the benchmarks?

richardpringle:
> Looks good to me.
> 
> Out of pure interest, can you check if this actually improves any of the benchmarks?

I'm not quite sure I understand what you're asking. `const` shouldn't change the behaviour at all, it will only make those functions available in `const` context. In order to improve a benchmark, you would have to change the code at the call-site. 

You should also be aware that removing `const` from a `pub const` function is a breaking change. Given the extensive use of the builder-pattern, this shouldn't be a concern, though.

PR #686, opened 2023-12-14
Implement Encode & Decode for Wrapping<T> types

mzachar:
Structures can clearly indicate that the value may overflow and it is intended behaviour


VictorKoenders:
I double checked serde's implementation (which matches your implementation, so you're good) and I noticed they also implement `std::cmp::Reverse<T>`. This is also a simple wrapper around `T`. Would you mind adding that as well?

mzachar:
I don't mind at all :)

Issue #647, opened 2023-06-10
impl Enode for IpAddr and the other like it ( unstable feature `ip_in_core` )


elpiel:
It's still awaiting stabilisation but it would be nice to include this in `2.0` release and track it's progress as an unstable feature.

Tracking issue: https://github.com/rust-lang/rust/issues/108443

VictorKoenders:
I'm not keen on having a dedicated feature for this. It shouldn't be a breaking change to modify this in a minor version, so once this is stabilized I think we can just patch it in

elpiel:
Ok, that's fair. My main pain point right now is on `no_std`. Right now I must create and maintain a separate struct for encoding/decoding IP addresses instead of using bincode's impl.
For no_std with e.g. embassy we're already using nightly so enabling unstable features like this one makes sense.

VictorKoenders:
As far as I can tell this should be stable in 1.76 next week

VictorKoenders:
Looks like I was wrong and this will be stable in rust 1.77, which should be released on march 21th

elpiel:
`ip_in_core` is stabilised!  :tada:

PR #694, opened 2024-01-28
impls for Ipv4Addr/Ipv6Addr

fox0:


fox0:
>error[E0658]: use of unstable library feature 'ip_in_core'

VictorKoenders:
Looks like that might be stable in 1.76 next week

VictorKoenders:
#647

PR #698, opened 2024-03-06
Add LICENSE.md to derive/

jfsulliv:
Adds a symlink from derive/LICENSE.md back to the LICENSE.md file in the root of the repository.  Since bincode_derive is published as a separate crate, this ensures the license is bundled appropriately with the crate.


VictorKoenders:
Broken on another issue, not going to block on that

PR #702, opened 2024-03-15
Update spec for `Option<T>` encoding

### LINKED ISSUE #700: Discriminant size in `fixint` serialization of `Option<T>`
I've run into an interesting corner case in the `bincode` specification:
- `Option<T>` is implemented as an `enum`
- The specification says that in `fixint` encoding, `enum` discriminants are always encoded as a `u32` ([source](https://github.com/bincode-org/bincode/blob/trunk/docs/spec.md#fixintencoding))
- However, a simple test program shows that an `Option<T>` is serialized with a 1-byte tag (instead of a `u32`)

```rust
fn main() {
    let i: Option<u32> = Some(100);
    let out = bincode::serialize(&i).unwrap();
    println!("{out:?}");

    let i: Option<u32> = None;
    let out = bincode::serialize(&i).unwrap();
    println!("{out:?}");

    let i: Result<u32, u32> = Ok(101);
    let out = bincode::serialize(&i).unwrap();
    println!("{out:?}");

    let i: Result<u32, u32> = Err(102);
    let out = bincode::serialize(&i).unwrap();
    println!("{out:?}");
}
```

```console
$ cargo run -q
[1, 100, 0, 0, 0]
[0]
[0, 0, 0, 0, 101, 0, 0, 0]
[1, 0, 0, 0, 102, 0, 0, 0]
```

Note that the `Result<T, E>` uses a 4-byte tag, as we'd expect from the specification.

I don't think the behavior should change at this point, but it would be good to reflect this corner case in the specification.  If that makes sense, I'd be happy to submit a PR to that effect.

mkeeter:
Fixes #700

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -86,7 +86,7 @@ assert_eq!(encoded.as_slice(), &[
 ]);
 ```
 
-## Options
+### Options
 `Option<T>` is always serialized using a single byte for the discriminant, even in `Fixint` encoding (which normally uses a `u32` for discriminant).
 
 ```rust
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -86,6 +86,23 @@ assert_eq!(encoded.as_slice(), &[
 ]);
 ```
 
+## Options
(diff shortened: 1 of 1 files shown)


VictorKoenders:
Thanks!

PR #705, opened 2024-03-17
Add (optional) smallvec decode/encode support

freqmod:
I added optional support for the smallvec crate which supports vectors inside a struct without a separate memory allocation. I tried implementing this with the newtype pattern, but i could not figure out a way to deserialise a without going via a vector (and thus allocating the memory that the crate was designed to avoid). 

The code is 90% copy paste from the code for (De)coding vectors. However I had to remove the optimisation for deserialising u8 as SmallVec don't use generics for the array size.

I hope that this change can be merged, and am happy to work with you if there are any changes that are needed in order to do so.

We probably also need to update the list of features in the docs. I can do that if necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,11 +29,13 @@ default = ["std", "derive"]
 std = ["alloc", "serde?/std"]
 alloc = ["serde?/alloc"]
 derive = ["bincode_derive"]
+smallvec = ["dep:smallvec"]
 
 [dependencies]
 bincode_derive = { path = "derive", version = "2.0.0-rc.3", optional = true }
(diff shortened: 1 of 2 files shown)


adamreichold:
> SmallVec don't use generics for the array size.

They do have a [`const_generics`](https://docs.rs/smallvec/latest/smallvec/#const_generics) feature.

VictorKoenders:
See https://github.com/bincode-org/bincode/pull/612 for previous work and discussion on this

freqmod:
I see, thanks for linking the issue for me. It is a bit strange that I didn't see this issue when I searched in github. 
I guess I will have to keep a fork around for now then until bincode v2 is released.

PR #708, opened 2024-03-20
Fixed #707

### LINKED ISSUE #707: Decode is only implemented for Arc<str> and not Rc<str>
https://github.com/bincode-org/bincode/blob/trunk/src/features/impl_alloc.rs#L454

vrtgs:
implemented `Decode` and `BorrowDecode` for Rc<str>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@
 *.swo
 .cargo
 .vscode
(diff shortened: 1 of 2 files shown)


VictorKoenders:
Thanks!

PR #715, opened 2024-05-28
Put clarifying language in migration guide regarding the serde:: and non-serde:: paths

mcclure:
This past week I ported a crate from bincode1 to bincode2. I [had some difficulty](https://github.com/bincode-org/bincode/issues/714) because I took the "migrate serde to bincode-derive" path, which was not appropriate for my application. I was following the migration guide, but I misread it.

I would like to recommend some minor language tweaks to make the migration guide more resistant to misinterpretation. Proposed language is attached, but I don't care if you use my specific language. I do suggest you make *some* form of the following changes:

* Make it clearer that the "Migrating with `serde`" and "Migrating to `bincode-derive`" sections describe *exclusive* options.
* Describe *why* someone might want to intentionally migrate with serde.

This would have helped me do my porting correctly the first time.

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -40,7 +40,9 @@ Because of confusion with `Options` defaults in bincode 1, we have made `Configu
 
 ## Migrating with `serde`
 
-Make sure to include bincode 2 with the `serde` feature enabled.
+You may wish to stick with `serde` when migrating to bincode 2, for example if you are using serde-exclusive derive features such as `#[serde(deserialize_with)]`.
+
+If so, make sure to include bincode 2 with the `serde` feature enabled.
 
 ```toml
 [dependencies]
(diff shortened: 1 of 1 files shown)


docs/migration_guide.md
```diff
@@ -40,7 +40,9 @@ Because of confusion with `Options` defaults in bincode 1, we have made `Configu
 
 ## Migrating with `serde`
 
-Make sure to include bincode 2 with the `serde` feature enabled.
+You may wish to stick with `serde` when migrating to bincode 2, for example if you are using serde-exclusive derive features such as `#[serde(deserialize_with)]`.
+
+If so, make sure to include bincode 2 with the `serde` feature enabled.
```
VictorKoenders: Might be useful to also mention that people should use the `bincode::serde::*` functions instead of the `bincode::*` functions

VictorKoenders (changes requested):

mcclure:
Nothing has changed, so this problem is still real. I still think this patch should be merged.

VictorKoenders:
Please fix the change I've requested and we can get this merged

mcclure:
Sorry, I misunderstood your comment before. Is that better?

VictorKoenders:
Thanks!

PR #723, opened 2024-08-22
Rewrite: replace the generated code in 'enc/impl_tuples' with macro

VendettaReborn:
## Why

The `Decode` trait is handwritten for tuples in `enc/impl_tuples`, and can be easily refactor by macro to reduce boilerplate code

## How

Added a macro `impl_tuple!` trait implementations for tuples.
Implemented the `Encode` trait for tuples with 1 to 16 elements using the macro.

## Test

the code can be compiled after the change, and the expanded code is the same as the previous one


VictorKoenders:
Sorry for the late response.

I'm hesitant to merge this because in my experience, a lot of people do not understand macros enough and it makes people less likely to contribute to an open source codebase. Also from preliminary testing, using a macro like this slightly increases initial build time for bincode.

What would be the upsides of having this in a macro? I don't think there's much of a problem with the currently existing boilerplate

VendettaReborn:
ok, i understand, i was just reading the code, and found the difference between encoder and decoder - the decoder utilized macro to make the code cleaner, so i made this change very quickly. It's true that the usage of macro will increase the building time, but from my viewpoint, it's not that consistent between the principle of readability and efficiency. 
anyway, it would still be fine with the current code. not a critical PR

PR #726, opened 2024-10-21
Update example in readme.md

samoylovfp:
Outdated example was not compiling, updated

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -31,33 +31,30 @@ library.
 ## Example
 
 ```rust
-use bincode::{config, Decode, Encode};
+use bincode::Options;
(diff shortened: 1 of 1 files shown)


xaocon:
@samoylovfp , I'm not an expert on this repo but I'm pretty sure that the examples are correct. I suspect the issue is that you added the newest stable version which is based on 1.x and the example in the default branch is for 2.x versions.

If I'm correct about this it might be worthwhile to add one of those "These docs are for a prerelease version, find stable here" kind of marks to the top until 2 is released.

VictorKoenders:
@xaocon is correct, the examples in the readme are for bincode 2.0-rc

PR #727, opened 2024-10-25
Reader: use BufRead instead of BufReader

realchonk:
This makes it, so any `BufRead` implementation can be used with `Reader`, not just `BufReader`.
This is very useful for my [fuse-ufs](https://github.com/realchonk/fuse-ufs) project.

The `&mut` implementation was removed, because [`BufRead` already provides one](https://doc.rust-lang.org/std/io/trait.BufRead.html#impl-BufRead-for-%26mut+B).

### DIFF
--- a/src/de/read.rs
+++ b/src/de/read.rs
@@ -31,26 +31,6 @@ pub trait Reader {
     fn consume(&mut self, _: usize) {}
 }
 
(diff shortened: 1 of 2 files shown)


realchonk:
Sorry, I'm an idiot. I just realized `decode_from_std_read` exists.

PR #728, opened 2024-10-25
Add prerelease warning to readme.md

xaocon:
It will be helpful to point users of the library to code examples that will work for them.

I started off with a `cargo add` and tried to use the example in the readme as guidance and ran into problems. When I started digging in to see the problem I saw #726 which seems to have had similar expectations.


VictorKoenders:
Thanks! Let's hope people read it 😆

PR #734, opened 2024-12-03
Implement basic traits for `Compat` and `BorrowCompat`

skibon02:
Wrapper types `Compat` and `BorrowCompat` do not introduce any kind of complexity or usage context to the inner type, and their only purpose is to provide an implementation for `bincode` traits.

So, it would be nice to make them as flexible as possible by deriving all common traits:
```rust
#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
```


VictorKoenders:
Makes a lot of sense, thanks!

PR #742, opened 2025-01-30
2.0.0 stable

VictorKoenders:


derive/Cargo.toml
```diff
@@ -1,11 +1,12 @@
 [package]
 name = "bincode_derive"
-version = "2.0.0-rc.3" # remember to update bincode
+version = "2.0.0" # remember to update bincode
 authors = [
     "Zoey Riordan <zoey@dos.cafe>",
     "Victor Koenders <bincode@trangar.com>",
 ]
 edition = "2021"
```
Havunen, line 8: Rust 1.85 brings support for Edition2024, maybe that could be updated as well?

VictorKoenders: We currently don't see the point of upgrading to 2024, as the rust compiler will still be able to run 2021 code. We do have the option to bump to 2024 later if we want to start adding new 2024 features

lherman-cs:
@VictorKoenders 2024 edition has become stable, https://github.com/rust-lang/rust/pull/133349. What can we do to help releasing bincode 2.0.0?

VictorKoenders:
Start by banning 40 hour work weeks, they're exhausting

VictorKoenders:
Would be really nice if we can merge #710 before releasing 2.0

ghost:
The readme's MSRV section should be updated to reflect the selected MSRV. Additionally, it should have a sentence mentioning that some features (e.g. `serde`) will change MSRV guarantees. I suggest adding a column in the lib.rs feature table documenting if a feature is MSRV affecting or not

PR #749, opened 2025-03-03
Decode context

ghost:
Supersedes #710. Changes are rebased on current trunk and build issues are fixed.

### DIFF
--- a/derive/src/attribute.rs
+++ b/derive/src/attribute.rs
@@ -1,4 +1,3 @@
-use proc_macro::Literal;
 use virtue::prelude::*;
 use virtue::utils::{parse_tagged_attribute, ParsedAttribute};
 
--- a/derive/src/attribute.rs
+++ b/derive/src/attribute.rs
@@ -1,3 +1,4 @@
+use proc_macro::Literal;
 use virtue::prelude::*;
 use virtue::utils::{parse_tagged_attribute, ParsedAttribute};
 
@@ -5,6 +6,7 @@ pub struct ContainerAttributes {
     pub crate_name: String,
     pub bounds: Option<(String, Literal)>,
     pub decode_bounds: Option<(String, Literal)>,
+    pub decode_context: Option<(String, Literal)>,
     pub borrow_decode_bounds: Option<(String, Literal)>,
     pub encode_bounds: Option<(String, Literal)>,
 }
@@ -15,6 +17,7 @@ impl Default for ContainerAttributes {
             crate_name: "::bincode".to_string(),
             bounds: None,
             decode_bounds: None,
+            decode_context: None,
             encode_bounds: None,
             borrow_decode_bounds: None,
         }
@@ -51,6 +54,15 @@ impl FromAttribute for ContainerAttributes {
                     let val_string = val.to_string();
                     if val_string.starts_with('"') && val_string.ends_with('"') {
                         result.decode_bounds =
+                            Some((val_string[1..val_string.len() - 1].to_string(), val));
+                    } else {
+                        return Err(Error::custom_at("Should be a literal str", val.span()));
+                    }
+                }
+                ParsedAttribute::Property(key, val) if key.to_string() == "decode_context" => {
+                    let val_string = val.to_string();
+                    if val_string.starts_with('"') && val_string.ends_with('"') {
+                        result.decode_context =
                             Some((val_string[1..val_string.len() - 1].to_string(), val));
                     } else {
                         return Err(Error::custom_at("Should be a literal str", val.span()));
--- a/src/de/decoder.rs
+++ b/src/de/decoder.rs
@@ -20,28 +20,30 @@ use crate::{config::Config, error::DecodeError, utils::Sealed};
 /// // this u32 can be any Decode
 /// let value = u32::decode(&mut decoder).unwrap();
 /// ```
-pub struct DecoderImpl<R, C: Config, Ctx> {
+pub struct DecoderImpl<R, C: Config, Context> {
     reader: R,
     config: C,
     bytes_read: usize,
-    ctx: Ctx,
+    context: Context,
 }
 
-impl<R: Reader, C: Config, Ctx> DecoderImpl<R, C, Ctx> {
+impl<R: Reader, C: Config, Context> DecoderImpl<R, C, Context> {
     /// Construct a new Decoder
-    pub const fn new(reader: R, config: C, ctx: Ctx) -> DecoderImpl<R, C, Ctx> {
+    pub const fn new(reader: R, config: C, context: Context) -> DecoderImpl<R, C, Context> {
         DecoderImpl {
             reader,
             config,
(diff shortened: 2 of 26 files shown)


derive/src/attribute.rs
```diff
@@ -56,6 +58,15 @@ impl FromAttribute for ContainerAttributes {
                         return Err(Error::custom_at("Should be a literal str", val.span()));
                     }
                 }
+                ParsedAttribute::Property(key, val) if key.to_string() == "decode_context" => {
```
VictorKoenders, line 61: We should start documenting the attributes on the derive macros. But we don't have any documentation so I don't think this PR should be blocked on this

src/de/decoder.rs
```diff
@@ -16,41 +16,51 @@ use crate::{config::Config, error::DecodeError, utils::Sealed};
...
+            context,
         }
     }
 }
 
-impl<R, C: Config> Sealed for DecoderImpl<R, C> {}
+impl<R, C: Config, Context> Sealed for DecoderImpl<'_, R, C, Context> {}
 
-impl<'de, R: BorrowReader<'de>, C: Config> BorrowDecoder<'de> for DecoderImpl<R, C> {
+impl<'de, R: BorrowReader<'de>, C: Config, Context> BorrowDecoder<'de>
```
VictorKoenders, line 45: I wonder if we can make the `Context: ?Sized`

src/de/mod.rs
```diff
@@ -71,38 +76,66 @@ pub use self::decoder::DecoderImpl;
...
-pub trait Decode: Sized {
+///
+/// You can use `Context` to require contexts for decoding a type:
+/// ```
+/// # /// # use bumpalo::Bump;
+/// use bincode::de::Decoder;
+/// use bincode::error::DecodeError;
+/// struct BytesInArena<'a>(bumpalo::collections::Vec<'a, u8>);
+/// impl<'a> bincode::Decode<&'a bumpalo::Bump> for BytesInArena<'a> {
+/// fn decode<D: Decoder>(decoder: &mut D) -> Result<Self, DecodeError> {
```
VictorKoenders, line 98: Formatting: need 2 spaces indent

src/de/mod.rs
```diff
@@ -71,38 +76,66 @@ pub use self::decoder::DecoderImpl;
...
+    fn decode<D: Decoder<Context = Context>>(decoder: &mut D) -> Result<Self, DecodeError>;
 }
 
 /// Trait that makes a type able to be decoded, akin to serde's `Deserialize` trait.
 ///
 /// This trait should be implemented for types that contain borrowed data, like `&str` and `&[u8]`. If your type does not have borrowed data, consider implementing [Decode] instead.
 ///
 /// This trait will be automatically implemented if you enable the `derive` feature and add `#[derive(bincode::Decode)]` to a type with a lifetime.
-pub trait BorrowDecode<'de>: Sized {
+pub trait BorrowDecode<'de, Context>: Sized {
```
VictorKoenders, line 113: Should we add a `'context` lifetime here? I'm honestly not sure

src/de/mod.rs
```diff
@@ -71,38 +76,66 @@ pub use self::decoder::DecoderImpl;
...
+/// use bincode::de::Decoder;
+/// use bincode::error::DecodeError;
+/// struct BytesInArena<'a>(bumpalo::collections::Vec<'a, u8>);
+/// impl<'a> bincode::Decode<&'a bumpalo::Bump> for BytesInArena<'a> {
+/// fn decode<D: Decoder>(decoder: &mut D) -> Result<Self, DecodeError> {
+///         todo!()
+///     }
+/// # }
+/// ```
+pub trait Decode<Context>: Sized {
```
VictorKoenders, line 103: Should we add a `'context` lifetime here? I'm honestly not sure

src/features/serde/de_owned.rs
```diff
@@ -24,28 +24,37 @@ impl<DE: Decoder> OwnedSerdeDecoder<DE> {
...
+impl<'r, 'context, C: Config, R: std::io::Read, Context>
+    OwnedSerdeDecoder<DecoderImpl<'context, IoReader<&'r mut R>, C, Context>>
+{
     /// Creates the decoder from an `std::io::Read` implementor.
     pub fn from_std_read(
         src: &'r mut R,
         config: C,
-    ) -> OwnedSerdeDecoder<DecoderImpl<IoReader<&'r mut R>, C>>
+        context: &'context mut Context,
+    ) -> OwnedSerdeDecoder<DecoderImpl<'context, IoReader<&'r mut R>, C, Context>>
```
VictorKoenders: Might be nice to leave this function context-less and add a new `from_std_read_with_context` function

src/features/serde/de_owned.rs
```diff
@@ -24,28 +24,37 @@ impl<DE: Decoder> OwnedSerdeDecoder<DE> {
...
+impl<'context, C: Config, R: Reader, Context>
+    OwnedSerdeDecoder<DecoderImpl<'context, R, C, Context>>
+{
     /// Creates the decoder from a [`Reader`] implementor.
-    pub fn from_reader(reader: R, config: C) -> OwnedSerdeDecoder<DecoderImpl<R, C>>
+    pub fn from_reader(
+        reader: R,
+        config: C,
+        context: &'context mut Context,
+    ) -> OwnedSerdeDecoder<DecoderImpl<'context, R, C, Context>>
```
VictorKoenders: Might be nice to leave this function context-less and add a new `from_reader_with_context` function

src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![no_std]
-#![warn(missing_docs, unused_lifetimes)]
+#![warn(unused_lifetimes)]
```
VictorKoenders: I think we should keep the `warn(missing_docs)` and add basic documentations to the new `_with_context` functions saying that they're the counterpart of the other functions, but with a context

src/de/mod.rs
ghost: Looking at the code, I'm not sure why we are passing `&mut Context` at all. Perhaps @branchseer can shed some light on that decision?

src/features/serde/de_owned.rs
ghost: Isn't possible so long as the decoders only contain a mutable reference to the context.

src/de/decoder.rs
ghost: No because the DecoderImpl owns a value of `Context`. Making it unsized would make weird problems. Unless we made it references all the way up.

ghost: Nvm, I'm dumb. I was looking at a local change that removed that :sigh:

derive/src/attribute.rs
ghost: Lets do this in another PR

VictorKoenders (approved):
lgtm ship it

PR #773, opened 2025-04-01
Add feature flag for `defmt` derives for errors

thelenlucas:
Added a new new non-default feature flag `defmt`, adds an optional dependency for [defmt](https://crates.io/crates/defmt), a lightweight logging crate for embedded machines, allowing for debugging over RTT. 

I dislike mutually exclusive features, but it might be worth adding here, as the `std` error's do not implement `defmt::Format`, as `defmt` is not intended to be used with the standard library. As such, they are wrapped with `debug`, which is inefficient, but unlikely to occur, due to the prior point. If we really wanted to break it out though, it should be trivial to do so via `compile_error!`.


VictorKoenders:
Sorry we haven't had the time to discuss this yet, life has been busy

VictorKoenders:
Couple of concerns we have:

defmt has a MSRV of 1.76, but we don't know what their MSRV policy is. Bincode has a strict 1.85 MSRV and does not want to bump this in the future for bincode 2.

Also it seems to not work with the `serde` feature enabled

PR #780, opened 2025-05-08
Implement core::error::Error for error types.

rosslannen:
With the error trait being moved from `std` to `core` in rust 1.81, implementing `core::error::Error` allows the error trait to be used in `[no_std]` environments for these types.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -63,6 +63,27 @@ impl core::fmt::Display for EncodeError {
     }
 }
 
+impl core::error::Error for EncodeError {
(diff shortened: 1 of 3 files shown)


VictorKoenders:
Looks like we'll want to remove the serde implementation in that case, otherwise looks good to me

rosslannen:
@VictorKoenders I got those removed, let me know if there's anything else I need to take care of.

VictorKoenders:
Thanks!

Issue #784, opened 2025-06-18
Allow a `MAX_SIZE` constant trait and generation with `derive`


roeishimron:
# Motivation: 
For allocating buffers in memory-lite environments and for automated security checks (like in #764), having a constant that supplies a reasonable upper-bound for encoded data is useful. Also, while decoding with an untrusted input, bounding the entire `decoder` will not prevent "playing around" with sizes of the internal fields of the encoded data, which may lead to unexpected behavior. In addition, having a bound that also validates the `Encoding` in alongside the `Decoding` is expected to catch logical bugs earlier thus reducing the development time. Finally, it is often unclear what is the correct global limit for the `with_limit` method, leading to redundant trial-and-error. 

# Intuition:
For most types, their maximum size is defined compile-time (like `u32` is maximum 4 bytes or `Enum` is the maximum of it's variant's maxima + discriminant size). However, collections with variable size are by-definition unblocked thus in order to provide the upped-bound mentioned above, the user must provide additional data, like the maximum *length* of the collection (and not its size).

## API Example for Getting the Maximum Size Without Collections:
```rust
#[derive(bincode::MaxSize]
pub struct TestMaxSize {
    b: u32,
    c: u32,
}

#[test]
fn test_() {
    assert_eq!(TestMaxSize::MAX_ENCODED_SIZE, 4+4);
}
```
# Implementation Suggestion:
Like [Postcard's](https://docs.rs/postcard/latest/src/postcard/max_size.rs.html) `max_size` feature, an implementation for that should likely include a trait called `MaxSize` with a `const MAX_ENCODED_SIZE: usize` property. Such trait could be implemented differently for every type. The `derive` feature will allow automatic creation of this trait's implementation, similar to the already existing `derive` macros.

## Possible Solutions For Communicating the Maximum Length of a Collection:
1.  Create a `newtype` generic for its maximum length while re-implementing `Encode`,  `Decode` and `MaxSize`  to also assert that the current length is less or equal to the maximum length. This solution is very simple but it is dirty because the access for this type must always be followed by `.0`.
2. Add an attribute providing the maximum length if required. This is more complicated because it will require macro-generated length-checking during `encode` and `decode`, but this solution provides a more natural API because it encapsulates the "encoding related" configuration in the macro-attribute world instead of the typing system.   

In order to avoid editing the internal Encoding and Decoding logic, `derive_decode` and `derive_decode`  could utilize the combination of 1,2 described above with: 
[...]

PR #786, opened 2025-06-25
Solve #784

roeishimron:


VictorKoenders:
Would you mind re-opening this PR on <https://github.com/bincode-org/bincode_contrib>? Let me know if you need any help.

roeishimron:
Cool idea! 
Just to make sure, `contib` is expected to `pub use bincode::*`  + additional features?

Note that in this PR there are some *changes* regarding `bincode` in addition to the new code introduced.
The *changes* are mainly:
1. `impl_alloc` until line 316: introducing common encoding using iterable (instead of re-implementation)
2.  `enc/mod.rs`, `de/mod.rs` , `lib.rs:226` : sharing the `SliceLenEncoding` type between encode and decode
3. `attribute.rs` and partially `derive_*.rs`: separating the user-given attributes from the serialization manner they require (it is an extension but I think it would be required at some point anyway)

Please let me know how (and if) you would like to split  those specific changes between `bincode` and `bincode-contrib`

VictorKoenders:
Our idea of `bincode_contrib` is to add commonly requested features that we don't feel like are a good fit in bincode itself. It is not expected to re-export everything in base bincode.

My idea for implementation is not make a new derive macro in `bincode_contrib_derive` that adds functionality on top of decode.

The implementation would use a combination of an `BincodeExactSize` trait, and a custom context or decoder to validate the size requirements at runtime. 

I'm hoping this can be done fully in `bincode_contrib`. If you run into issues, let me know.

roeishimron:
It can be done in `contrib` but I don't see how it can avoid additional derive macro for implementing the `MaxSize` trait.
Moreover, without the internal changes inserted at `impl_alloc` the alternative decoder would have to re-implement `decode` and `borrow_decode` for all collections.

roeishimron:
> Our idea of `bincode_contrib` is to add commonly requested features that we don't feel like are a good fit in bincode itself. It is not expected to re-export everything in base bincode.
> 
> My idea for implementation is not make a new derive macro in `bincode_contrib_derive` that adds functionality on top of decode.
> 
> The implementation would use a combination of an `BincodeExactSize` trait, and a custom context or decoder to validate the size requirements at runtime.
> 
> I'm hoping this can be done fully in `bincode_contrib`. If you run into issues, let me know.

Also, how would such repo "sync" version-wise with the original `bincode` if its not requiring it (thus not aware of the user's `bincode` version)? 
Since `bincode-contrib` is an extention for `bincode`, as a user I'd expect it to support `bincode`'s API breaks and (I think that) this could be done only by re-exporting `bincode`.

VictorKoenders:
> support bincode's API breaks

That will not happen until bincode 3.0, which will not happen for a long time.

However it is a contrib crate, if people want it to be exposed then that can be added

derive/src/attribute.rs
```diff
@@ -8,6 +8,7 @@ pub struct ContainerAttributes {
     pub decode_context: Option<(String, Literal)>,
     pub borrow_decode_bounds: Option<(String, Literal)>,
     pub encode_bounds: Option<(String, Literal)>,
+    pub maxsize_bounds: Option<(String, Literal)>,
```
LielHarel: If I understand correctly (I may have mistakes since I don't feel good at proc macro yet) - `max_bounds` will be set with a value here if we use `#[bincode(maxsize_bounds = "100")]` on our struct or enum, right? I didn't find in tests an example for this.
I ask because it will help me to understand more.

src/lib.rs
```diff
@@ -221,6 +222,9 @@ pub fn decode_from_reader<D: de::Decode<()>, R: Reader, C: Config>(
     D::decode(&mut decoder)
 }
 
+/// The underlying type for decoding and encoding slice len
+pub type SliceLenEncoding = u64;
```
LielHarel: :1st_place_medal:

derive/src/attribute.rs
roeishimron: No. There's a not-very-documented feature that supports passing generic bounds to the trait implementations. It looks like
```rust
#[derive(MaxSize)]
#[bincode(maxsize_bounds = "MyTrait")]
struct A<T: MaxSize>{
a: T
}
```
Should result  an implementation like
```rust
impl<T: MyTrait> MaxSize for A<T>{...}
```
Instead of the expected
```rust
impl<T: MaxSize> MaxSize for A<T>{...}
```

roeishimron: As shown in tests, the attribute you refer looks like
```rust
        #[derive(bincode::MaxSize, bincode::Encode, bincode::Decode)]
        pub(crate) enum TestEnumSerializationAlloc {
            u32,
            #[bincode(maxlen = 5)]
            V(Vec<u8>),
            TestEnumMaxSizeInnerStruct,
        }
```
And parsed with `SerializationKind` addition.

PR #791, opened 2025-08-14
Improve compile times by replacing ouroboros with self_cell

Voultapher:
Tested on a AMD 5900X.

Before:

```
$ hyperfine -p "cargo clean" "cargo test --no-run"
  Time (mean ± σ):      8.085 s ±  0.072 s    [User: 44.474 s, System: 8.612 s]
  Range (min … max):    7.985 s …  8.221 s    10 runs
```

After:

```
$ hyperfine -p "cargo clean" "cargo test --no-run"
  Time (mean ± σ):      7.504 s ±  0.059 s    [User: 42.242 s, System: 8.075 s]
  Range (min … max):    7.392 s …  7.589 s    10 runs
```

self_cell is a widely used crate and mature crate.


VictorKoenders:
This changes the semantics of the test. I'd be okay with adding self_cell as a new test if you want to change that.

